PR #6, opened 2015-05-29
Fix edge case where isSystem() will report erroneus results

jviotti:


lib/linux.coffee
```diff
@@ -31,4 +31,5 @@ exports.isSystem = (drive, callback) ->
 			return callback(false)
 
 		result = tableParser.parse(stdout)
-		return callback(result[0].RM?[0] isnt '1')
+		rmFlag = result[0].RM?[0] or result[0]['MAJ:MIN'][1]
```
Page-: Why are you looking at the `MAJ:MIN` property for whether it's removable?

jviotti: @Page- There seems to be an edge case where the table parser module I'm using interprets the RM value as the second value of MAJ:MIN. I'll add a comment explaining that.

Page-: Have you/can you report this upstream?

jviotti: @Page- Reported to `table-parser`.

PR #8, opened 2015-06-30
Change windows wmic diskdrive to wmic logicaldisk

acabelloj:
I have changed the windows call 'wmic diskdrive' to use 'logicaldisk' as it is more useful returning DeviceID 'C:' instead of '\\\\.\\PHYSICALDRIVE0'. Also removed build directory as it is auto-generated.

Right now if you want to read a file from a USB using nodejs you will need to pass the logical volume name for windows:

````javascript
var fs = require('fs');
var filename = "D:\\file.json";
fs.readFile(filename, function(err, data) {
    if (err) throw err;
    console.log(data);
  });
````

If drive list returns in windows:
````javascript
[ { device: 'C:', description: 'Windows 8.1', size: '661 GB' },
  { device: 'D:', description: 'USB1', size: '15 GB' } ]
````
instead of:
````javascript
{
 device: '\\\\.\\PHYSICALDRIVE0',
 description: 'WDC WD10JPVX-75JC3T0',
 size: '661 GB'
},
````
We will be able to identify and read/write files from mounted devices.

### CODE REVIEW

**Comment by Page-** [comment]
The two are actually different use cases, this is an example from my machine:
```
>wmic diskdrive get DeviceID, Caption, Size
Caption                       DeviceID            Size
SAMSUNG SSD PM810 2.5" 7      \\.\PHYSICALDRIVE0  256052966400
JMCR SD/MMC SCSI Disk Device  \\.\PHYSICALDRIVE1  15554004480
```
and
```
>wmic logicaldisk get DeviceID, Size, VolumeName
DeviceID  Size          VolumeName
C:        256052928512  OS
D:
```

`diskdrive` lists all my disks, including my unmounted sd card, meaning it is possible to address and write to it directly at the disk level (eg for burning images).  
`logicaldisk` however does not show it (due to it being unmounted and hence not having a disk letter), it does however show my (empty) optical drive, which would allow accessing mounted drives on a filesystem level (eg in order to read a file)

What I would say is that these should be different methods, as both have their uses, rather than one being strictly better than the other

acabelloj:
I agree, better split up to keep both functionalities.

jviotti:
The benefit of `drivelist` is that the exact same interface is implemented for all major operating systems. Adding a new function specifically for Windows will break this benefit, and from a user standpoint, I have to manually check if I'm running Windows before calling this particular function.

@acabelloj I suggest that you create another module that lists logical drives (since the use case is different), or you create a module that takes `drivelist` output as an input, and translates physical drive ids to logical names.

acabelloj:
Thanks @jviotti, I think I'm opting for forking the module as translating the physical drive ids is going to be harder.

jviotti:
@acabelloj Sounds great, let us know how it goes. Good luck! :+1:

PR #13, opened 2015-08-10
Replace wmic with custom vbscript in Windows

mkorakakis:


lib/win32.coffee
```diff
@@ -3,7 +3,7 @@ _ = require('lodash')
 tableParser = require('table-parser')
 
 exports.list = (callback) ->
-	childProcess.exec 'wmic diskdrive get DeviceID, Caption, Size', {}, (error, stdout, stderr) ->
+	childProcess.exec 'cscript ./shell_scripts/win_drivers.vbs //Nologo', {}, (error, stdout, stderr) ->
```
jviotti: Can you use `__dirname` and `path.join()` for safety?

jviotti:
Can you rename the `shell_scripts` directory to just `scripts`?

build/win32.js
```diff
@@ -1,30 +1,32 @@
...
 _ = require('lodash');
 
-tableParser = require('table-parser');
+path = require('path');
 
 exports.list = function(callback) {
-  return childProcess.exec('wmic diskdrive get DeviceID, Caption, Size', {}, function(error, stdout, stderr) {
-    var result;
+  var script;
+  script = path.join(__dirname, '..', '/scripts/win_drives.vbs');
```
jviotti: Can you split `/scripts/win_drives.vbs` into separate arguments to `path.join`? The purpose is to not hardcode any file path delimiter, and let `path.join` use the appropriate one.

lib/win32.coffee
```diff
@@ -1,25 +1,28 @@
...
+	childProcess.exec "cscript #{script} //Nologo", {}, (error, stdout, stderr) ->
 		return callback(error) if error?
 
 		if not _.isEmpty(stderr)
 			return callback(new Error(stderr))
 
-		result = tableParser.parse(stdout)
+		output = stdout.trim()
+		output = output.replace(/\r/, '')
+		output = output.split(/\n/g)
```
jviotti: Can you to `trim` and `replace` all in this line for simplicity?

```coffee
output = stdout.trim().replace(/\r/, '').split(/\n/g)
```

lib/win32.coffee
```diff
@@ -1,25 +1,28 @@
...
 
-			if row.DeviceID.length > 1
-				row.Caption = row.Caption.concat(_.initial(row.DeviceID))
+			size = _.parseInt(drive_info[2]) / 1e+9 or undefined
 
 			return {
-				device: _.last(row.DeviceID)
-				description: row.Caption.join(' ')
+				device: drive_info[1],
+				description: drive_info[0].trim(),
```
jviotti: For safety, we could trim all the elements of `drive_info` with a `_.map`.

lib/win32.coffee
```diff
@@ -1,25 +1,28 @@
...
 
-		result = tableParser.parse(stdout)
+		output = stdout.trim()
+		output = output.replace(/\r/, '')
+		output = output.split(/\n/g)
 
-		result = _.map result, (row) ->
-			size = _.parseInt(row.Size?[0]) / 1e+9 or undefined
+		result = _.map output, (row) ->
+			drive_info = row.split('\t')
```
jviotti: Can you use camelCase for the variable name?

lib/win32.coffee
mkorakakis: Like this?
```			
driveInfo = _.map driveInfo, (element) ->
   element.trim()
```

build/win32.js
mkorakakis:
```
script = path.join(__dirname, '..', 'scripts', 'win_drives.vbs')
```

lib/win32.coffee
jviotti: Yeah, I like to add an explicit `return` statement in the trim line since we care about the return value.

PR #42, opened 2015-11-23
Execute scripts with execFile

jviotti:
`execFile` is patched by Electron to be able to execute scripts inside
`asar` packages (see https://github.com/atom/electron/issues/3512).

Using `exec`, as we have been doing until now, causes `drivelist` to
throw `ENOENT` on the scripts.

Collateral changes:

- We removed the double quotes between darwin and linux paths to
workaround errors when paths contain spaces.

(see https://github.com/resin-io/drivelist/issues/39).

This is no longer needed when using `execFile` (in fact, this function
throws an error when quoting the path). This was tested manually to
prevent a regression.

- We create an intermediary `win32.bat` script that calls
`win32.vbs` since `execFile` knows how to call this type of file
automatically.

### DIFF
--- a/build/scripts.js
+++ b/build/scripts.js
@@ -11,13 +11,15 @@ path = require('path');
 scriptsPath = path.join(__dirname, '..', 'scripts');
 
 exports.paths = {
-  win32: "cscript \"" + (path.join(scriptsPath, 'win32.vbs')) + "\" //Nologo",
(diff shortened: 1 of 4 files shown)


jviotti:
Tested in OS X.

jviotti:
Tested in Windows 10.

jviotti:
Tested in Ubuntu.

PR #47, opened 2016-01-01
Trim trailing commas in parser

jviotti:
On GNU/Linux, we emit a comma-separated string of the different
mountpoints of a certain drive, for example `/mnt,/tmp/mnt,`.

For simplicity, the `linux.sh` script replaces all `\n` with a comma,
causing all mountpoints to have a trailing comma, independently of
having a single mountpoint or many.

Fixes: https://github.com/resin-io/drivelist/issues/46

### DIFF
--- a/build/parse.js
+++ b/build/parse.js
@@ -51,6 +51,11 @@ module.exports = function(input) {
     if (_.isString(result)) {
       return _.object([result], [null]);
     }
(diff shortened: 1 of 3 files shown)


tborychowski:
:+1: that fixes my issue, thanks!

PR #48, opened 2016-01-01
Fix boolean values being surrounded by quotes

jviotti:
Instead of `false` or `true` we get `'false'` and `'true'` respectively.


jviotti:
This regression was introduced in https://github.com/resin-io/drivelist/issues/46

PR #49, opened 2016-01-04
Implement name property

jviotti:
This property represents the "display name" of a drive. In the case of
OS X and GNU/Linux, this is simply the device name (e.g: `/dev/disk2`)
while on Windows it represents the drive letter (e.g: `C:`).

The purpose of this distinction is to abstract the way end users
identify drives, since something like `\\.\PhysicalDriveN` makes no
sense for a Windows user.


jviotti:
Tested in all supported OSes.

PR #57, opened 2016-03-22
Use udev to determine if a drive is removable on GNU/Linux

jviotti:


scripts/linux.sh
```diff
@@ -22,12 +22,12 @@ for disk in $DISKS; do
...
-  removable=`lsblk -d $device --output RM | ignore_first_line | trim`
-
-  if [[ "$removable" == "1" ]]; then
+  eval "`udevadm info --query=property --export --export-prefix=UDEV_ --name=$disk`"
+  if [[ "`lsblk -d $device --output RM | ignore_first_line | trim`" == "1" ]] || \
+     [[ "$UDEV_ID_DRIVE_FLASH_SD" == "1" ]] || \
+     [[ "$UDEV_ID_DRIVE_MEDIA_FLASH_SD" == "1" ]] || \
+     [[ "$UDEV_ID_BUS" == "usb" ]]
+  then
     echo "system: False"
```
agherzan: This error seems way to generic. Can you make it more detailed / descriptive? Same for True.

scripts/linux.sh
```diff
@@ -22,12 +22,12 @@ for disk in $DISKS; do
...
-  # by checking the removeable flag.
-  # There might be a better way in GNU/Linux systems.
-  removable=`lsblk -d $device --output RM | ignore_first_line | trim`
-
-  if [[ "$removable" == "1" ]]; then
+  eval "`udevadm info --query=property --export --export-prefix=UDEV_ --name=$disk`"
+  if [[ "`lsblk -d $device --output RM | ignore_first_line | trim`" == "1" ]] || \
+     [[ "$UDEV_ID_DRIVE_FLASH_SD" == "1" ]] || \
+     [[ "$UDEV_ID_DRIVE_MEDIA_FLASH_SD" == "1" ]] || \
+     [[ "$UDEV_ID_BUS" == "usb" ]]
```
agherzan: So you return False when is removable?

scripts/linux.sh
jviotti: What do you mean by error? I'm logging a property here. The output from this script is then parsed as a JavaScript object.

scripts/linux.sh
jviotti: I consider `system` drives to be non-removable drives. I got the "system" naming convention from somewhere else, I think another NodeJS module, but I see it causes some confusion. Maybe I can rename to `removable`?

agherzan: Uh. With this explanation this makes sense. You can either rename it or state this in a comment or so.

scripts/linux.sh
agherzan: Got it. I though it was just pure logging. All good

scripts/linux.sh
jviotti: Cool, I'm making a note to rename this property in a new pull request.

PR #60, opened 2016-03-25
Fixes Win32.bat reporting incorrect system drive

userabuser:
Issue #61

### DIFF
--- a/scripts/win32.bat
+++ b/scripts/win32.bat
@@ -14,7 +14,7 @@ Set objWMIService = GetObject("winmgmts:\\" & strComputer & "\root\cimv2")
 
 Set colDiskDrives = objWMIService.ExecQuery("SELECT * FROM Win32_DiskDrive")
 
-Set colOperatingSystems = objWMIService.ExecQuery ("Select SystemDrive from Win32_OperatingSystem")
+Set colOperatingSystems = objWMIService.ExecQuery ("SELECT SystemDrive FROM Win32_OperatingSystem")
 
 For Each objOperatingSystem in colOperatingSystems
     On Error Resume Next
--- a/scripts/win32.bat
+++ b/scripts/win32.bat
@@ -16,12 +16,13 @@ Set colDiskDrives = objWMIService.ExecQuery("SELECT * FROM Win32_DiskDrive")
 
 Set colOperatingSystems = objWMIService.ExecQuery ("SELECT SystemDrive FROM Win32_OperatingSystem")
 
+Err.Clear
 For Each objOperatingSystem in colOperatingSystems
     On Error Resume Next
     '' get the OS System Drive if exists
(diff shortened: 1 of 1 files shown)


scripts/win32.bat
```diff
@@ -5,12 +5,27 @@ exit /b
...
 Set colDiskDrives = objWMIService.ExecQuery("SELECT * FROM Win32_DiskDrive")
 
+Set colOperatingSystems = objWMIService.ExecQuery ("SELECT SystemDrive FROM Win32_OperatingSystem")
+
+For Each objOperatingSystem in colOperatingSystems
+    On Error Resume Next
+    '' get the OS System Drive if exists
+    OSDrive = objOperatingSystem.Properties_("SystemDrive")   
+    If Err.Number <> 0 Then
+        OSDrive = Err.Description
```
jviotti: So if `objOperatingSystem.Properties_("SystemDrive")` fails, we assign `OSDrive` to `Err.Description`? Why is this? As I understand your intention was to set something that will always differ from `objLogicalDisk.DeviceID`, so `system: False` is printed when an error happens right? If so, maybe worth clarifying the intention in a comment, and potentially set `OSDrive` to something more obvious to the eye, like `false` (not sure if VBScript has the concept of `undefined`, `null`?)

scripts/win32.bat
```diff
@@ -5,12 +5,27 @@ exit /b
...
+        Err.Clear
```
jviotti: Maybe worth calling `Err.Clear` *before* the loop starts as well, since if an error exists before the loop, `OSDrive` will not get the right value, and the drive will be reported as `system: false`.

scripts/win32.bat
userabuser: Correct..

Originally I had `OSDrive = False`, but during my various tests, somehow reverted back to `Err.Description`.

The original intention behind `Err.Description` was to provide a meaningful error to `system: {value}` because there'd be no other way to debug the case where `SystemDrive` was not set as a property. However that's probably over-engineering.

I agree that `OSDrive = False` is probably sufficient enough.

scripts/win32.bat
userabuser: Good catch, my VBScript kungfu is mediocre so I would not have necessarily thought of that had you not mentioned it.

userabuser:
@jviotti improved `For Each` loop over `colOperatingSystems` based on discussion. See: https://github.com/resin-io-modules/drivelist/pull/60/commits/79e4811f3ee1126519415dd366c72e461edf9213

jviotti:
@userabuser Cheers! Looks good to merge code-wise. I'll give it a quick spin in my PC and merge!

jviotti:
@userabuser Merged. Thanks again for the PR! I published v2.0.10 to NPM with your fix.

userabuser:
@jviotti No problem... Thank you for the great tool. Very useful.

Issue #72, opened 2016-04-24
[Windows] Show drives without any Windows partitions


peakwinter:
Presently on Windows, this module only shows drives that have Windows mountable partitions (i.e. drives that show up with drive letters after `Win32_LogicalDiskToPartition`). Empty removable drives, or removable drives that only have HFS or ext partitions, will not show up. Is this intended behaviour or just an oversight? If the latter I would be happy to submit a pull request to fix...

Thanks!

jviotti:
Hi @peakwinter ,

If there is a way to make them visible, happy to take a PR :)

peakwinter:
I have a commit in a local branch that works and am ready to send a pull -- a question first though, what should be the value of the `name` property in these cases? If it's not a mountable drive, it will have a device ID (like `\\.\PHYSICALDRIVE1`) but not a drive letter, of course. Name could also be set to `null` but that seems a bit fishy to me. And would likely require a change to what is displayed for `name` in all cases...

jviotti:
Hi @peakwinter ,

Sounds exciting, looking forward to give it a go! :)

Regarding `name`, it already defaults to `null` when the drive doesn't have a drive letter attached to it, so `null` should be good to go :+1:

jviotti:
Thanks a lot @peakwinter! I published v3.1.0 to NPM with your addition.

peakwinter:
No prob, thanks for the great module :)

PR #73, opened 2016-04-26
[Windows] Show non-mountable disk drives

peakwinter:
This commit returns disk drives that do not contain any mountable partitions under Windows, useful for seeing drives with HFS, ext* or other types of partitions.

(See #72)

### DIFF
--- a/scripts/win32.bat
+++ b/scripts/win32.bat
@@ -29,7 +29,7 @@ Next
 
 For Each objDrive In colDiskDrives
     strDeviceID = Replace(objDrive.DeviceID, "\", "\\")
-    foundLDs = 0
+    containsLogicalDisk = False
     Set colPartitions = objWMIService.ExecQuery _
         ("ASSOCIATORS OF {Win32_DiskDrive.DeviceID=""" & _
             strDeviceID & """} WHERE AssocClass = " & _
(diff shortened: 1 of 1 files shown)


scripts/win32.bat
```diff
@@ -39,6 +40,7 @@ For Each objDrive In colDiskDrives
                 objPartition.DeviceID & """} WHERE AssocClass = " & _
                     "Win32_LogicalDiskToPartition")
         For Each objLogicalDisk In colLogicalDisks
+            foundLDs = foundLDs + 1
```
jviotti: Given that we only care if at least one logical disk was found, and not how many, I recommend this variable to be a boolean one instead, called `containsLogicalDisk` or something like that.

If we reach to this point, we can set to `True`, and then compare with `False` below.

scripts/win32.bat
```diff
@@ -54,5 +56,14 @@ For Each objDrive In colDiskDrives
             Wscript.Echo ""
         Next
     Next
+    If foundLDs = 0 Then
+      Wscript.Echo "device: """ & Replace(objDrive.DeviceID, "\", "\\") & """"
+      Wscript.Echo "description: """ & objDrive.Caption & """"
+      Wscript.Echo "size: " & objDrive.Size
+      Wscript.Echo "mountpoint: Null"
+      Wscript.Echo "name: Null"
+      Wscript.Echo "system: False"
```
jviotti: I believe we should default to `system: True` here, or we should find a way to better detect this. As it is now, a program using this module, might detect any non-mountable disk drive as a removable one.

peakwinter: Hm, that's a tricky one. As it is, though, the only drive marked as system is the one the OS is running on. Other non-removable drives are still designated as `system: false`.

Another option would be to set it as `null` I guess?

jviotti: If we can't infer the value, I rather default it to `True`, so we're on the safe side.

jviotti: @peakwinter Can you also please include a commend above this line saying that we can't infer its value, so we default to `True` to be on the safe side?

peakwinter: Done!

PR #84, opened 2016-07-29
Qmarks can occur in block device descriptions

Fordi:
This breaks the YAML when it occurs.  Added escaping to descriptions.

### DIFF
--- a/scripts/linux.sh
+++ b/scripts/linux.sh
@@ -26,7 +26,8 @@ for disk in $DISKS; do
   fi
 
   device="/dev/$disk"
-  description=`lsblk -d $device --output MODEL | ignore_first_line`
+  description="$(lsblk -d $device --output MODEL | ignore_first_line | trim)"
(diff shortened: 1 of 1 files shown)


jviotti:
Hi @Fordi ,

Thanks a lot for this patch. I can confirm it works fine (I mocked `lsblk` returning a description with question marks, which fails without your patch).

Do you mind explain how this works? I honestly have no idea about it :)

jviotti:
I've released this in v3.2.5.

Fordi:
Sure.  I was talking quotation marks, but question marks, spaces, and possibly exclamation points would have broken too.

The first change just wraps the run-and-capture in bash quotes, which ensures `description` gets the whole content every time.

The added line makes use of bash's built-in regular expression substitution to precede "s with a backslash to escape it for YAML's sake.

jviotti:
Very nice. Bash's built-in regular expression substitution was the piece I was missing. Thanks a lot!

Issue #88, opened 2016-08-20
Description on El Capitain


eemi2010:
I m getting WDC WD5000LPVX-22V0TT0 for the description instead of a readable name. What I m  I doing wrong.

jviotti:
Hi @eemi2010 ,

We only display what the OS detects to be the description. Can you show us the output of `diskutil info /dev/diskN`, where N corresponds to your drive id, to see if we can find such information from other place?

eemi2010:
Hey there. This is what  I get when  I run diskutil info /dev/disk1. Hope it can help.

Device Identifier:        disk1
   Device Node:              /dev/disk1
   Whole:                    Yes
   Part of Whole:            disk1
   Device / Media Name:      WDC WD5000LPVX-22V0TT0

   Volume Name:              Macintosh HD

   Mounted:                  Yes
   Mount Point:              /

   File System Personality:  Journaled HFS+
   Type (Bundle):            hfs
   Name (User Visible):      Mac OS Extended (Journaled)
   Journal:                  Journal size 40960 KB at offset 0x15be5000
   Owners:                   Enabled

   Content (IOContent):      Apple_HFS
   OS Can Be Installed:      Yes
   Recovery Disk:            disk0s3
   Media Type:               Generic
   Protocol:                 SATA
   SMART Status:             Not Supported
   Volume UUID:              611FF4C1-1528-35A1-B45F-11C80D8A852D
   Disk / Partition UUID:    E9BADD4A-BCED-4B1F-B7BC-AB534B049BB2

   Total Size:               498.9 GB (498876809216 Bytes) (exactly 974368768 512-Byte-Units)
   Volume Free Space:        129.9 GB (129917546496 Bytes) (exactly 253745208 512-Byte-Units)
   Device Block Size:        512 Bytes
   Allocation Block Size:    4096 Bytes

   Read-Only Media:          No
   Read-Only Volume:         No

   Device Location:          Internal
   Removable Media:          No

   Solid State:              No
   Virtual:                  Yes
   OS 9 Drivers:             No
   Low Level Format:         Not supp

jviotti:
Hey there!

What you see is indeed what the OS reports as the drive name:

```
Device / Media Name:      WDC WD5000LPVX-22V0TT0
```

Can you provide an example of the string you were expecting instead?

eemi2010:
Hi, 

[...]

PR #89, opened 2016-08-22
Attempt to display the volume name as description in OS X

jviotti:
The `Volume Name` key contains a much more human friendly description
for the drive. This PR attempts to use is over `Device / Media Name`
unless its not applicable. For example

```
Volume Name:              Not applicable (no file system)
```

Fixes: https://github.com/resin-io-modules/drivelist/issues/88

### DIFF
--- a/scripts/darwin.sh
+++ b/scripts/darwin.sh
@@ -15,6 +15,7 @@ for disk in $DISKS; do
 
   device=`echo "$diskinfo" | get_key "Device Node"`
   description=`echo "$diskinfo" | get_key "Device / Media Name"`
+  volume_name=`echo "$diskinfo" | get_key "Volume Name"`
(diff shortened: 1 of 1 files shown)


dustMason:
It turns out that in my app, the description (Device / Media Name) value is more important than the Volume Name. This changes causes that value to be lost!

My app is a companion to a synthesizer that loads sounds by acting as a USB storage device. The volume name is generic (NO NAME) but the Device / Media Name contains the product name (OP-1). I need this device name in order to properly detect it.

Not sure what the proper fix for this is. Let me know if you have an idea and I will create a PR. In the meantime I'll be using a fork in my project.

Thanks!

jviotti:
Hm, what about joining the volume name and the device name separated by a hyphen if they are both available? Would that work for you?

dustMason:
Sounds good! See #166

Issue #75, opened 2016-05-09
OS X: No `mountpoint` for USB devices


mikew:
The OS X script ends up running `diskutil info /dev/disk2`, when it should be running `diskutil info /dev/disk2s1`. Using something like this is more reliable, but a bit slower

```bash
list_disks () {
	mounts=$(mount)
	diskutil list | while read line; do
		id=$(echo "$line" | rev | cut -d ' ' -f 1 | rev)
		if [[ "$id" == disk* ]]; then
			if [[ "$mounts" == *"/dev/$id on"* ]]; then
				echo "/dev/$id"
			fi
		fi
	done
}

for disk in $(list_disks); do
  # ...
```

mikew:
Another way to write it to avoid both `mount` and `diskutil list` is:

```bash
list_disks () {
	mount | while read line; do
		dev=$(echo "$line" | cut -d ' ' -f 1)
		if [[ "$dev" == /dev/disk* ]]; then
			echo "$dev"
		fi
	done
}
```

jviotti:
Hi @mikew ,

Your code means that we'll be outputting information about each partition as a separate object in the final array. The purpose of `drivelist` is to detect the connected drives (without caring about the partitions), so that's the reason we only call `diskutil info` for top level disks.

I'd be happy to take a PR that recursively scans the partitions of a disk, and creates a comma separated list of mount points for the top level disk.

mikew:
The first function checks agains the device in the output of `mount`.

Before:

```
$ ./node_modules/drivelist/scripts/darwin.sh
device: /dev/disk0
description: APPLE SSD SM0256G
size: 251000193024
mountpoint:
name: /dev/disk0
system: True

device: /dev/disk1
description: APPLE SSD SM0256G
size: 199000412160
mountpoint: /
name: /dev/disk1
system: True

device: /dev/disk2
description: Ultra Fit
size: 62176362496
mountpoint:
name: /dev/disk2
system: False
```

After:

```
$ ./node_modules/drivelist/scripts/darwin.sh
device: /dev/disk1
description: APPLE SSD SM0256G
size: 199000412160
mountpoint: /
name: /dev/disk1
system: True

device: /dev/disk0s4
description: BOOTCAMP
size: 50778341376
mountpoint: /Volumes/BOOTCAMP
name: /dev/disk0s4
system: True

device: /dev/disk2s1
description: Untitled 1
size: 34359738368
mountpoint: /Volumes/ESD-USB
[...]

PR #93, opened 2016-08-25
Add 'mountpoints' property on Mac

joerick:
As per #75, this adds a `mountpoints` property, which is an array of the mount points of the partitions of the devices connected.

Example output:

```
$ node -e "require('./build/drivelist').list(function(err, list) { console.log(list) })"                 master!
[ { device: '/dev/disk0',
    description: 'APPLE SSD SM0512F',
    size: 500277790720,
    mountpoint: null,
    name: '/dev/disk0',
    raw: '/dev/rdisk0',
    protected: false,
    system: true,
    mountpoints: [ '/', '/Volumes/BOOTCAMP' ] },
  { device: '/dev/disk2',
    description: 'SD Card Reader',
    size: 7969177600,
    mountpoint: null,
    name: '/dev/disk2',
    raw: '/dev/rdisk2',
    protected: false,
    system: false,
    mountpoints: [ '/Volumes/boot', '/Volumes/Untitled' ] },
  { device: '/dev/disk4',
    description: 'External USB 3.0',
    size: 1000204883968,
    mountpoint: null,
    name: '/dev/disk4',
    raw: '/dev/rdisk4',
    protected: false,
    system: true,
    mountpoints: [] },
  { device: '/dev/disk5',
    description: 'Joe backup HD',
    size: 999489667072,
    mountpoint: '/Volumes/Joe backup HD 1',
    name: '/dev/disk5',
    raw: '/dev/rdisk5',
    protected: false,
    system: true,
    mountpoints: [ '/Volumes/Joe backup HD 1' ] } ]
```

Worth noting that this does increase the execution time to 5 seconds on my machine, compared to 2 seconds before this patch. Not a concern for my applications though.

### DIFF
--- a/scripts/darwin.sh
+++ b/scripts/darwin.sh
@@ -8,6 +8,10 @@ function get_until_paren {
   awk 'match($0, "\\(|$"){ print substr($0, 0, RSTART - 1) }'
 }
 
+function get_device_identifiers {
+  awk 'match($0, /disk[0-9]+(s[0-9]+)?$/, arr){ print "/dev/" arr[0] }'
+}
+
 DISKS="`diskutil list | grep '^\/' | get_until_paren`"
 
 for disk in $DISKS; do
@@ -61,6 +65,24 @@ for disk in $DISKS; do
   else
     echo "system: False"
   fi
+
+  partitions="`diskutil list $disk | get_device_identifiers`"
+  partitionmountpoints=""
+
+  for partition in $partitions; do 
+    partitioninfo="`diskutil info $partition`"
+
+    partitionmountpoint=`echo "$partitioninfo" | get_key "Mount Point"`
+
+    if [ "$partitionmountpoint" ]; then
+      partitionmountpoints=$partitionmountpoints,$partitionmountpoint
+    fi
+  done
+  
+  # trim leading ,
+  partitionmountpoints=${partitionmountpoints#,}
+
+  echo "mountpoints: [ $partitionmountpoints ]"
 
   echo ""
 done
--- a/scripts/darwin.sh
+++ b/scripts/darwin.sh
@@ -26,7 +26,7 @@ for disk in $DISKS; do
   location=`echo "$diskinfo" | get_key "Device Location"`
   size=`echo "$diskinfo" | sed 's/Disk Size/Total Size/g' | get_key "Total Size" | perl -n -e'/\((\d+)\sBytes\)/ && print $1'`
 
-  mountpoint=`echo "$mount_output" | perl -n -e'm{'"^${disk}(s[0-9]+)? on (.*) \(.*\)$"'} && print ",$2"'`
+  mountpoint=`echo "$mount_output" | perl -n -e'm{^'"${disk}"'(s[0-9]+)? on (.*) \(.*\)$} && print ",$2"'`
   # trim leading ,
   mountpoint=${mountpoint#,}
 
--- a/scripts/darwin.sh
+++ b/scripts/darwin.sh
@@ -13,6 +13,7 @@ function get_device_identifiers {
 }
 
 DISKS="`diskutil list | grep '^\/' | get_until_paren`"
+mount_output="`mount`"
 
 for disk in $DISKS; do
   diskinfo="`diskutil info $disk`"
@@ -24,11 +25,14 @@ for disk in $DISKS; do
 
   description=`echo "$diskinfo" | get_key "Device / Media Name"`
   volume_name=`echo "$diskinfo" | get_key "Volume Name"`
-  mountpoint=`echo "$diskinfo" | get_key "Mount Point"`
   removable=`echo "$diskinfo" | get_key "Removable Media"`
   protected=`echo "$diskinfo" | get_key "Read-Only Media"`
   location=`echo "$diskinfo" | get_key "Device Location"`
   size=`echo "$diskinfo" | sed 's/Disk Size/Total Size/g' | get_key "Total Size" | perl -n -e'/\((\d+)\sBytes\)/ && print $1'`
+
+  mountpoint=`echo "$mount_output" | perl -n -e'm{'"^${disk}(s[0-9]+)? on (.*) \(.*\)$"'} && print ",$2"'`
+  # trim leading ,
+  mountpoint=${mountpoint#,}
 
   # Omit mounted DMG images
   if [ "$description" == "Disk Image" ]; then
@@ -65,24 +69,6 @@ for disk in $DISKS; do
   else
     echo "system: False"
   fi
-
-  partitions="`diskutil list $disk | get_device_identifiers`"
-  partitionmountpoints=""
-
-  for partition in $partitions; do 
-    partitioninfo="`diskutil info $partition`"
-
-    partitionmountpoint=`echo "$partitioninfo" | get_key "Mount Point"`
-
-    if [ "$partitionmountpoint" ]; then
-      partitionmountpoints=$partitionmountpoints,$partitionmountpoint
-    fi
-  done
-  
-  # trim leading ,
-  partitionmountpoints=${partitionmountpoints#,}
-
-  echo "mountpoints: [ $partitionmountpoints ]"
 
   echo ""
 done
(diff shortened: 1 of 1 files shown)


jviotti:
Hey there!

Thanks a lot for the pull request. The `mountpoint` property is supposed to contain a comma separated list of mount points, so no need to create a new property. The fact that is `null` is a bug, so lets append those values to `mountpoint` instead.

Regarding the execution time, that is a problem, given that other applications run this functions every 2 seconds or so to poll for new devices, but I'm sure there are ways to optimise. I'll leave some comments on the source code itself.

scripts/darwin.sh
```diff
@@ -8,6 +8,10 @@ function get_until_paren {
   awk 'match($0, "\\(|$"){ print substr($0, 0, RSTART - 1) }'
 }
 
+function get_device_identifiers {
+  awk 'match($0, /disk[0-9]+(s[0-9]+)?$/, arr){ print "/dev/" arr[0] }'
```
jviotti: This throws some syntax errors in my system (El Capitan):

```
awk: syntax error at source line 1
 context is
        match($0, >>>  /disk[0-9]+(s[0-9]+)?$/, <<<
awk: bailing out at source line 1
```

scripts/darwin.sh
```diff
@@ -62,5 +66,23 @@ for disk in $DISKS; do
     echo "system: False"
   fi
 
+  partitions="`diskutil list $disk | get_device_identifiers`"
```
jviotti: Notice that the list of partitions that you get from `diskutil list <disk>` is a subset of the whole `diskutil list` output.

For example:

```
$ diskutil list
/dev/disk0 (internal, physical):
   #:                       TYPE NAME                    SIZE       IDENTIFIER
   0:      GUID_partition_scheme                        *251.0 GB   disk0
   1:                        EFI EFI                     209.7 MB   disk0s1
   2:          Apple_CoreStorage Macintosh HD            250.1 GB   disk0s2
   3:                 Apple_Boot Recovery HD             650.0 MB   disk0s3
/dev/disk1 (internal, virtual):
   #:                       TYPE NAME                    SIZE       IDENTIFIER
   0:                  Apple_HFS Macintosh HD           +249.8 GB   disk1
                                 Logical Volume on disk0s2
                                 DCD23031-6322-4269-A142-CD36C8FD95D7
                                 Unlocked Encrypted
/dev/disk2 (external, physical):
   #:                       TYPE NAME                    SIZE       IDENTIFIER
   0:     FDisk_partition_scheme                        *7.8 GB     disk2
   1:                 DOS_FAT_32 UNTITLED                7.8 GB     disk2s1
```
***
```
$ diskutil list /dev/disk2
/dev/disk2 (external, physical):
   #:                       TYPE NAME                    SIZE       IDENTIFIER
   0:     FDisk_partition_scheme                        *7.8 GB     disk2
   1:                 DOS_FAT_32 UNTITLED                7.8 GB     disk2s1
```

Notice that the `diskutil list /dev/disk2` output is just the bottom part of `diskutil list`. This means we could re-use the initial `diskutil list` call at line 15 somehow to also get the partitions to avoid an extra `diskutil` call.

scripts/darwin.sh
```diff
@@ -62,5 +66,23 @@ for disk in $DISKS; do
...
+
+    if [ "$partitionmountpoint" ]; then
+      partitionmountpoints=$partitionmountpoints,$partitionmountpoint
+    fi
+  done
+  
+  # trim leading ,
+  partitionmountpoints=${partitionmountpoints#,}
+
+  echo "mountpoints: [ $partitionmountpoints ]"
```
jviotti: Lets just expose a comma separated string at this point for consistency with the other operating systems. Happy to accept a PR that parses comma separated lists as arrays in https://github.com/resin-io-modules/drivelist/blob/master/lib/parse.coffee later.

scripts/darwin.sh
```diff
@@ -62,5 +66,23 @@ for disk in $DISKS; do
...
+
+    partitionmountpoint=`echo "$partitioninfo" | get_key "Mount Point"`
+
+    if [ "$partitionmountpoint" ]; then
+      partitionmountpoints=$partitionmountpoints,$partitionmountpoint
+    fi
+  done
+  
+  # trim leading ,
+  partitionmountpoints=${partitionmountpoints#,}
```
jviotti: If we move the array parsing code to the JS side of this module, it means that we can also take care of trimming there, which I believe is more convenient.

scripts/darwin.sh
```diff
@@ -62,5 +66,23 @@ for disk in $DISKS; do
     echo "system: False"
   fi
 
+  partitions="`diskutil list $disk | get_device_identifiers`"
+  partitionmountpoints=""
+
+  for partition in $partitions; do 
+    partitioninfo="`diskutil info $partition`"
```
jviotti: In order to avoid calling `diskutil info` every time, which is an expensive operation, I suggest saving the output of running `mount`, and re-using it in every loop to find the mount points:

```
mount_output="`mount`"
mountpoint=`echo "$mount_output" | grep "^/dev/disk2s1" | cut -d ' ' -f 3`
```

jviotti: BTW, there are some very cool suggestions on the original issue about how to list available drives, including partitions. In particular, there is one using just `mount`, which means we can cache `mount`'s output at the very beginning and use it for everything.

joerick:
Okay, great feedback and yes, if performance is an issue I think the `mount` output is the way to go.

joerick:
- Now calling `mount` at the start and parsing that output for each disk.
- The script looks for lines with /dev/diskX or /dev/diskXsX and records those mountpoints.
- Switched to perl and somehow avoided the dreaded [LTS](https://en.wikipedia.org/wiki/Leaning_toothpick_syndrome).
- Changed the output to put all mountpoints, comma-separated under the `mountpoint` label.

Here's how it looks:

```
joerick@joerick ~/02 Tingbot/drivelist> time node -e "require('./build/drivelist').list(function(err, list) { console.log(list) })"            master!
[ { device: '/dev/disk0',
    description: 'APPLE SSD SM0512F',
    size: 500277790720,
    mountpoint: '/,/Volumes/BOOTCAMP',
    name: '/dev/disk0',
    raw: '/dev/rdisk0',
    protected: false,
    system: true },
  { device: '/dev/disk2',
    description: 'SD Card Reader',
    size: 7969177600,
    mountpoint: '/Volumes/boot,/Volumes/Untitled',
    name: '/dev/disk2',
    raw: '/dev/rdisk2',
    protected: false,
    system: false },
  { device: '/dev/disk4',
    description: 'External USB 3.0',
    size: 1000204883968,
    mountpoint: null,
    name: '/dev/disk4',
    raw: '/dev/rdisk4',
    protected: false,
    system: true },
  { device: '/dev/disk5',
    description: 'Joe backup HD',
    size: 999489667072,
    mountpoint: '/Volumes/Joe backup HD 1',
    name: '/dev/disk5',
    raw: '/dev/rdisk5',
    protected: false,
    system: true } ]
node -e   0.45s user 0.38s system 48% cpu 1.728 total
```

jviotti:
Hi @joerick ,

Amazing progress! I still didn't look at the code, but can spot some things in the example you posted above:

- `/dev/disk1` is not listed at all, despite being in `mount`'s output: `/dev/disk1s1 on /Volumes/Etcher`.
- `/dev/disk3` is not listed at all neither. `mount` has this entry: `/dev/disk3s1 on /Volumes/tide-electron`

jviotti:
The code looks great! With regarding to my previous comment, perhaps you ran `drivelist` after removing the drives I raised up as missing?

jviotti:
Ah, `/dev/disk1` is a "Disk Image" (a mounted DMG), right? In that case omitting is the correct behaviour :+1:

joerick:
Cheers! Yes we are releasing a RaspberryPi-based hardware product ([Tingbot](http://tingbot.com)) and we're recommending people flash the OS to an SD card using Etcher! Also needed a cross-platform way to list partitions to program some settings etc., so found this project. 👍

jviotti:
Cool stuff! Every Etcher discussion takes place in a public Gitter channel (https://gitter.im/resin-io/etcher), so feel free to ping me there if there is anything I can help with!

Thanks a lot for the PR, will merge and release a new version to NPM in a minute.

jviotti:
Checkout v3.3.2!

PR #105, opened 2016-10-31
Syntax typo

lurch:
Quote the Mac OS X example raw values.

I don't have a Mac with which to test, but I'm _assuming_ this update is correct, based on the examples shown for the other OSes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Mac OS X:
     description: 'GUID_partition_scheme',
     size: 68719476736,
     mountpoint: '/',
-    raw: /dev/rdisk0,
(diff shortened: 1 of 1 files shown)


jviotti:
Definitely right, good catch!

lurch:
Does the same change need to be made to `doc/README.hbs` ? (why are there two README files?)

jviotti:
Ah, right, I forgot this project is using the new README convention. So `doc/README.hbs` is a template that gets compiled with the current JSDoc annotations, so the README always as the latest docs.

jviotti:
I sent a PR fixing it myself.

PR #114, opened 2016-11-08
Linux: get `description` for SD card readers that appear as e.g. /dev/mmcblk0

lurch:
@jviotti We probably ought to check if anything similar happens on Windows and OS X?


jviotti:
Confirmed it works on both OS X and Windows as well!

PR #116, opened 2016-11-13
fix yaml array parsing error,(substr from F: to F)

afandiyusuf:
i thought i solve a problem with yaml parsing error,
that because the mountainpoint letter give you "F:" instead of "F"

so the result of your script is
mountainpoints:
    - path: F:

my script
mountainpoints:
    - path: F


jviotti:
Hey there,

Thanks a lot for the contribution, however I think we should preserve the colon, since its part of the actual path, and usually needed when you pass it to Windows I/O APIs.

As an alternative, I've surrounded the drive letter in double quotes, which should do the trick. Can you give it a go in your computer? https://github.com/resin-io-modules/drivelist/pull/117

afandiyusuf:
yeah, you are right

PR #119, opened 2016-11-15
bugfix for OSX drives with description ':', which leads to invalid yaml

### LINKED ISSUE #118: YAMLException: incomplete explicit mapping pair
The current version of drivelist gives me the following error:

    YAMLException: incomplete explicit mapping pair; a key node is missed at line 2, column 14:
        description: :
                     ^
        at generateError (/myproject/node_modules/drivelist/node_modules/js-yaml/lib/js-yaml/loader.js:162:10)
        at throwError (/myproject/node_modules/drivelist/node_modules/js-yaml/lib/js-yaml/loader.js:168:9)
        at readBlockMapping (/myproject/node_modules/drivelist/node_modules/js-yaml/lib/js-yaml/loader.js:995:9)
        at composeNode (/myproject/node_modules/drivelist/node_modules/js-yaml/lib/js-yaml/loader.js:1327:12)
        at readDocument (/myproject/node_modules/drivelist/node_modules/js-yaml/lib/js-yaml/loader.js:1489:3)
        at loadDocuments (/myproject/node_modules/drivelist/node_modules/js-yaml/lib/js-yaml/loader.js:1545:5)
        at load (/myproject/node_modules/drivelist/node_modules/js-yaml/lib/js-yaml/loader.js:1562:19)
        at Object.safeLoad (/myproject/node_modules/drivelist/node_modules/js-yaml/lib/js-yaml/loader.js:1580:10)
        at _.compact._.map (/myproject/node_modules/drivelist/lib/parse.js:89:25)
        at arrayMap (/myproject/node_modules/drivelist/node_modules/lodash/lodash.js:660:23)

It seems that the references version of `js-yaml` is at fault. Any idea?

metachris:
fixes #118

### DIFF
--- a/scripts/darwin.sh
+++ b/scripts/darwin.sh
@@ -28,6 +28,10 @@ for disk in $DISKS; do
 
   mountpoints=`echo "$mount_output" | perl -n -e'm{^'"${disk}"'(s[0-9]+)? on (.*) \(.*\)$} && print "$2\n"'`
 
+  if [ "$volume_name" == ":" ]; then
+      volume_name=""
+  fi
+
   # Omit mounted DMG images
   if [ "$description" == "Disk Image" ]; then
     continue


jviotti:
Hey there!

Thanks a lot for the pull request. What happens if a drive has a description that "contains" a colon? I think we should surround the value of `description` in double quotes, which will prevent YAML from complaining but will allow you to get the actual description, either if its just a colon, or a string containing a colon.

What do you think?

lurch:
Yeah, double-quoting the description sounds like a better option to me. AFAICT the `linux.sh` script already does that, so I'm surprised that `darwin.sh` doesn't?

jviotti:
> AFAICT the linux.sh script already does that, so I'm surprised that darwin.sh doesn't?

Yeah, they get out of sync very easily. Maybe we can do some linting or something to make sure they all get the same fixes (for example, another script that checks that every property printed by the script is surrounded in double quotes)?

Closing in favor of https://github.com/resin-io-modules/drivelist/pull/127

PR #120, opened 2016-11-16
Replace System.Collections.ArrayList with out own List implementation

jviotti:
`ArrayList` comes from the .NET framework, and if the framework is not
installed on the user's machine, or it unconfigured in any way, the
script will just bail out with a cryptic error.

See: https://github.com/resin-io/etcher/issues/870

### DIFF
--- a/scripts/win32.bat
+++ b/scripts/win32.bat
@@ -11,6 +11,26 @@ exit /b
 ----- Begin wsf script --->
 <job><script language="VBScript">
 
+Class List
+	Private Dictionary
+
+	Private Sub Class_Initialize()
+		Set Dictionary = CreateObject("Scripting.Dictionary")
+	End Sub
+
+	Public Sub Add(element)
+		Dictionary.Add element, ""
+	End Sub
+
+	Public Function GetArray()
+		GetArray = Dictionary.Keys()
+	End Function
+
+	Public Function Count
+		Count = UBound(Dictionary.Keys()) + 1
+	End Function
+End Class
+
 Set WMIService = GetObject("winmgmts:\\.\root\cimv2")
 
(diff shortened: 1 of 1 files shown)


lurch:
"with out own" in the title should be "with our own"
"or it unconfigured" should be "or is unconfigured"

scripts/win32.bat
```diff
@@ -107,8 +127,8 @@ For Each TopLevelDrive In GetTopLevelDrives()
 		Wscript.Echo "mountpoints: []"
 	Else
 		Wscript.Echo "mountpoints:"
-		For Each Mountpoint In TopLevelDrive.Item("Mountpoints")
-			Wscript.Echo "  - path: """ & Mountpoint & """"
+		For Each Mountpoint In TopLevelDrive.Item("Mountpoints").GetArray()
+			Wscript.Echo "	- path: """ & Mountpoint & """"
```
lurch: What's the change on this line?

lurch:
Does the use of a Dictionary mean the array doesn't have guaranteed order? Or are Dictionaries in VBScript different to dictionaries in other languages?

scripts/win32.bat
jviotti: Ugh, the space was replaced with a tab, which causes YAML to break :/ I'll fix in a second.

jviotti:
@lurch The order of the keys seems to be preserved when getting the array of keys with `Dictionary.Keys()`, but can't find any documentation that backs that up :/

scripts/win32.bat
jviotti: The thing is that the unit tests test the parser, but don't run the actual scripts (since we would need some way to mock connected devices, or something like that).

PR #129, opened 2016-12-21
Consider multiple "Device Location" keys in Mac Minis

jviotti:
Mac Minis might report multiple "Device Location" keys in the output of
`diskutil info /dev/diskN`, for example:

```
   Device Identifier:        disk1
   Device Node:              /dev/disk1
   Whole:                    Yes
   Part of Whole:            disk1
   Device / Media Name:      ST2000LM003 HN-M201RAD

   Volume Name:              Not applicable (no file system)
   Mounted:                  Not applicable (no file system)
   File System:              None

   Content (IOContent):      GUID_partition_scheme
   OS Can Be Installed:      No
   Media Type:               Generic
   Protocol:                 SATA
   SMART Status:             Verified

   Disk Size:                2.0 TB (2000398934016 Bytes) (exactly 3907029168 512-Byte-Units)
   Device Block Size:        512 Bytes

   Read-Only Media:          No
   Read-Only Volume:         Not applicable (no file system)

   Device Location:          Internal
   Removable Media:          Fixed

   Solid State:              No
   Virtual:                  No
   OS 9 Drivers:             No
   Low Level Format:         Not supported
   Device Location:          "Lower"
```

Notice that the above example contains:

```
   Device Location:          Internal
```

And:

```
   Device Location:          "Lower"
```

Which then causes equality checks against the value to fail (since the
final value retrieved by `get_key` contains multiple lines).

As a solution, we modify `get_key` to accumulate all found values into a
space separated string, and we then use `=~` to check against it.

See: https://github.com/resin-io/etcher/issues/983

### DIFF
--- a/scripts/darwin.sh
+++ b/scripts/darwin.sh
@@ -1,7 +1,17 @@
 #!/bin/sh
 
 function get_key {
-  grep "$1" | awk -F "  +" '{ print $3 }'
+  awk -F "  +" "
+    /${1/\//\/}/ {
+      if (values) {
+        values = values \" \" \$3
+      } else {
+        values = \$3
+      }
+    }
+
+    END { print values }
+  "
 }
 
 function get_until_paren {
@@ -64,7 +74,7 @@ for disk in $DISKS; do
 
   if [[ "$device" == "/dev/disk0" ]] || \
      [[ "$removable" == "No" ]] || \
-     [[ ( "$location" == "Internal" ) && ( "$removable" != "Yes" ) && ( "$removable" != "Removable" ) ]] || \
+     [[ ( "$location" =~ "Internal" ) && ( "$removable" != "Yes" ) && ( "$removable" != "Removable" ) ]] || \
      echo "$mountpoints" | grep "^/$"
   then
     echo "system: True"


lurch:
Not that it matters now, but http://stackoverflow.com/a/33540439 has a much neater method of joining multiple lines into one, than your complicated `awk` script ;-)

jviotti:
@lurch Do you mean the `cat` one? How would we do it here (without having a separate file)? I'm happy to replace the awk script with something simpler!

lurch:
Doing some local testing (see below) with dummy output, I believe it's simply enough to change the `get_key` function to
```bash
function get_key {
  echo $(grep "$1" | awk -F "  +" '{ print $3 }')
}
```

Let me know if you want me to submit a PR.


P.S. I tested by creating a `test_echo.sh` which says
```bash
#!/bin/bash
cat <<EOF
   Device Identifier:        disk1
   Device Node:              /dev/disk1
   Whole:                    Yes
   Part of Whole:            disk1
   Device / Media Name:      ST2000LM003 HN-M201RAD

   Volume Name:              Not applicable (no file system)
   Mounted:                  Not applicable (no file system)
   File System:              None

   Content (IOContent):      GUID_partition_scheme
   OS Can Be Installed:      No
   Media Type:               Generic
   Protocol:                 SATA
   SMART Status:             Verified

   Disk Size:                2.0 TB (2000398934016 Bytes) (exactly 3907029168 512-Byte-Units)
   Device Block Size:        512 Bytes

   Read-Only Media:          No
   Read-Only Volume:         Not applicable (no file system)

   Device Location:          Internal
   Removable Media:          Fixed

   Solid State:              No
   Virtual:                  No
   OS 9 Drivers:             No
   Low Level Format:         Not supported
   Device Location:          "Lower"
EOF
```
and a `test_join.sh` which says
```bash
#!/bin/bash
function get_key1 {
  grep "$1" | awk -F "  +" '{ print $3 }'
}
function get_key2 {
  awk -F "  +" "
    /${1/\//\/}/ {
      if (values) {
        values = values \" \" \$3
      } else {
        values = \$3
      }
    }
    END { print values }
  "
}
function get_key3 {
  echo $(grep "$1" | awk -F "  +" '{ print $3 }')
}

diskinfo="`~/test_echo.sh`"
location1=`echo "$diskinfo" | get_key1 "Device Location"`
echo "Loc1: #$location1#"
location2=`echo "$diskinfo" | get_key2 "Device Location"`
echo "Loc2: #$location2#"
location3=`echo "$diskinfo" | get_key3 "Device Location"`
echo "Loc3: #$location3#"
```

Running `test_join.sh` then outputs:
```
Loc1: #Internal
"Lower"#
Loc2: #Internal "Lower"#
Loc3: #Internal "Lower"#
```

jviotti:
Amazing!

> Let me know if you want me to submit a PR.

Yeah, please!

lurch:
Just for reference: simpler version of `get_key` got PRed in #131

PR #133, opened 2017-01-27
Double quote PATH declaration in win32.bat

jviotti:
Failing to do so can introduce problems when some of the directories in
include spaces in their paths, eventually manifesting themselves as
weird "is not recognized as an internal or external command,
operable program or batch file" errors.

See: https://github.com/resin-io/etcher/issues/1025

### DIFF
--- a/scripts/win32.bat
+++ b/scripts/win32.bat
@@ -3,7 +3,11 @@
 
 :: Ensure System32 is in the PATH, to avoid weird
 :: 'cscript' is not recognized as an internal or external command"" errors.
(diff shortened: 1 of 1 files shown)


lurch:
It's not the spaces in the path that's the problem, but having an `&` in the path - that's why this only showed up for MySQL (which adds `C:\Program Files (x86)\MySQL\MySQL Fabric 1.5.4 & MySQL Utilities 1.5.4 1.5` to the `%PATH%`) and not other paths / programs.

PR #134, opened 2017-01-30
Improve darwin.sh based on shellcheck tips

jviotti:
These were the addressed warnings:

```
In scripts/darwin.sh line 4:
  echo $(grep "$1" | awk -F "  +" '{ print $3 }')
       ^-- SC2046: Quote this to prevent word splitting.
       ^-- SC2005: Useless echo? Instead of 'echo $(cmd)', just use 'cmd'.

In scripts/darwin.sh line 11:
DISKS="`diskutil list | grep '^\/' | get_until_paren`"
       ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 12:
mount_output="`mount`"
              ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 15:
  diskinfo="`diskutil info $disk`"
            ^-- SC2006: Use $(..) instead of legacy `..`.
                           ^-- SC2086: Double quote to prevent globbing and word splitting.

In scripts/darwin.sh line 17:
  device=`echo "$diskinfo" | get_key "Device Node"`
         ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 20:
  raw_device=`echo "$device" | sed "s/disk/rdisk/g"`
             ^-- SC2006: Use $(..) instead of legacy `..`.
              ^-- SC2001: See if you can use ${variable//search/replace} instead.

In scripts/darwin.sh line 22:
  description=`echo "$diskinfo" | get_key "Device / Media Name"`
              ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 23:
  volume_name=`echo "$diskinfo" | get_key "Volume Name"`
              ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 24:
  removable=`echo "$diskinfo" | get_key "Removable Media"`
            ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 25:
  protected=`echo "$diskinfo" | get_key "Read-Only Media"`
            ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 26:
  location=`echo "$diskinfo" | get_key "Device Location"`
           ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 27:
  size=`echo "$diskinfo" | sed 's/Disk Size/Total Size/g' | get_key "Total Size" | perl -n -e'/\((\d+)\sBytes\)/ && print $1'`
       ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 29:
  mountpoints=`echo "$mount_output" | perl -n -e'm{^'"${disk}"'(s[0-9]+)? on (.*) \(.*\)$} && print "$2\n"'`
              ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/darwin.sh line 52:
    echo "$mountpoints" | while read mountpoint ; do
                                ^-- SC2162: read without -r will mangle backslashes.
```

### DIFF
--- a/scripts/darwin.sh
+++ b/scripts/darwin.sh
@@ -1,35 +1,43 @@
-#!/bin/sh
+#!/bin/bash
 
-function get_key {
+set -u
+set -e
+
+get_key() {
+
+  # This expression is intentionally unquoted so that
+  # multiple lines get joined as a single one.
+  # See https://github.com/resin-io-modules/drivelist/pull/129
   echo $(grep "$1" | awk -F "  +" '{ print $3 }')
+
 }
 
-function get_until_paren {
+get_until_paren() {
   awk 'match($0, "\\(|$"){ print substr($0, 0, RSTART - 1) }'
 }
 
-DISKS="`diskutil list | grep '^\/' | get_until_paren`"
-mount_output="`mount`"
+DISKS="$(diskutil list | grep '^\/' | get_until_paren)"
+mount_output="$(mount)"
 
 for disk in $DISKS; do
-  diskinfo="`diskutil info $disk`"
+  diskinfo="$(diskutil info "$disk")"
 
-  device=`echo "$diskinfo" | get_key "Device Node"`
+  device="$(echo "$diskinfo" | get_key "Device Node")"
 
   # See http://superuser.com/q/631592
-  raw_device=`echo "$device" | sed "s/disk/rdisk/g"`
+  raw_device="${device//disk/rdisk}"
 
-  description=`echo "$diskinfo" | get_key "Device / Media Name"`
-  volume_name=`echo "$diskinfo" | get_key "Volume Name"`
-  removable=`echo "$diskinfo" | get_key "Removable Media"`
-  protected=`echo "$diskinfo" | get_key "Read-Only Media"`
-  location=`echo "$diskinfo" | get_key "Device Location"`
-  size=`echo "$diskinfo" | sed 's/Disk Size/Total Size/g' | get_key "Total Size" | perl -n -e'/\((\d+)\sBytes\)/ && print $1'`
+  description="$(echo "$diskinfo" | get_key "Device / Media Name")"
+  volume_name="$(echo "$diskinfo" | get_key "Volume Name")"
+  removable="$(echo "$diskinfo" | get_key "Removable Media")"
+  protected="$(echo "$diskinfo" | get_key "Read-Only Media")"
+  location="$(echo "$diskinfo" | get_key "Device Location")"
+  size="$(echo "$diskinfo" | sed 's/Disk Size/Total Size/g' | get_key "Total Size" | perl -n -e'/\((\d+)\sBytes\)/ && print $1')"
 
-  mountpoints=`echo "$mount_output" | perl -n -e'm{^'"${disk}"'(s[0-9]+)? on (.*) \(.*\)$} && print "$2\n"'`
+  mountpoints="$(echo "$mount_output" | perl -n -e'm{^'"${disk}"'(s[0-9]+)? on (.*) \(.*\)$} && print "$2\n"')"
 
   # Omit mounted DMG images
-  if [ "$description" == "Disk Image" ]; then
+  if [[ "$description" == "Disk Image" ]]; then
     continue
   fi
 
@@ -45,11 +53,11 @@ for disk in $DISKS; do
 
   echo "size: $size"
 
-  if [ -z "$mountpoints" ]; then
+  if [[ -z "$mountpoints" ]]; then
     echo "mountpoints: []"
   else
     echo "mountpoints:"
-    echo "$mountpoints" | while read mountpoint ; do
+    echo "$mountpoints" | while read -r mountpoint ; do
       echo "  - path: $mountpoint"
     done
   fi


scripts/darwin.sh
```diff
@@ -1,35 +1,38 @@
...
+  protected="$(echo "$diskinfo" | get_key "Read-Only Media")"
+  location="$(echo "$diskinfo" | get_key "Device Location")"
+  size="$(echo "$diskinfo" | sed 's/Disk Size/Total Size/g' | get_key "Total Size" | perl -n -e'/\((\d+)\sBytes\)/ && print $1')"
 
-  mountpoints=`echo "$mount_output" | perl -n -e'm{^'"${disk}"'(s[0-9]+)? on (.*) \(.*\)$} && print "$2\n"'`
+  mountpoints="$(echo "$mount_output" | perl -n -e'm{^'"${disk}"'(s[0-9]+)? on (.*) \(.*\)$} && print "$2\n"')"
 
   # Omit mounted DMG images
-  if [ "$description" == "Disk Image" ]; then
+  if [ "$description" = "Disk Image" ]; then
```
lurch: You've switched to just a single-equals here...

scripts/darwin.sh
```diff
@@ -49,20 +52,20 @@ for disk in $DISKS; do
...
-    echo "$mountpoints" | while read mountpoint ; do
+    echo "$mountpoints" | while read -r mountpoint ; do
       echo "  - path: $mountpoint"
     done
   fi
 
   echo "raw: $raw_device"
 
-  if [[ "$protected" == "Yes" ]]; then
+  if [ "$protected" == "Yes" ]; then
```
lurch: ...but then left a double-equals here!

scripts/darwin.sh
```diff
@@ -49,20 +52,20 @@ for disk in $DISKS; do
...
     echo "protected: True"
   else
     echo "protected: False"
   fi
 
-  if [[ "$device" == "/dev/disk0" ]] || \
+  if [ "$device" = "/dev/disk0" ] || \
      [[ ( "$removable" == "No" ) || ("$removable" == "Fixed") ]] || \
      [[ ( "$location" =~ "Internal" ) && ( "$removable" != "Yes" ) && ( "$removable" != "Removable" ) ]] || \
```
lurch: **If** we're removing all usage of `[[` then we obviously need to update these lines too.

scripts/darwin.sh
```diff
@@ -1,35 +1,38 @@
-#!/bin/sh
+#!/bin/bash
 
-function get_key {
-  echo $(grep "$1" | awk -F "  +" '{ print $3 }')
+set -u
+set -e
+
+get_key() {
+  grep "$1" | awk -F "  +" '{ print $3 }'
```
lurch: You need to revert the change on this line - it's **deliberately** unquoted, see #131

scripts/darwin.sh
```diff
@@ -1,35 +1,38 @@
-#!/bin/sh
+#!/bin/bash
```
lurch: A lot of the shellcheck output was based on using POSIX sh instead of bash (as this script was using bash-style features), but you've now switched to bash here anyway!

So IMHO you should either leave it as `/bin/sh` and fix up all the shellcheck errors (and you need to re-run it after every change before committing) to make the script sh-compliant, OR you should revert all the changes you've made here, switch to bash, and then re-run shellcheck (which will then presumably then operate in bash-mode, and not point out all the POSIX sh errors). Hope that makes sense ;-)

jviotti: I did re-run shellcheck after changing to bash (sorry for the confusion of putting the initial warnings there. I'll update). shellcheck basically stopped request POSIX compliance stuff but didn't add anything new.

scripts/darwin.sh
jviotti: True. This was caused due to the shell change confusion. I left all of them as `==`.

scripts/darwin.sh
jviotti: shellcheck requires me to put `[[` in these cases because of `=~` and nested expressions (separated by parenthesis), but it requires me to switch to `[` for anything else.

scripts/darwin.sh
jviotti: True. Fixed! I added a comment as well.

scripts/darwin.sh
lurch: Not when it's operating in bash-mode it doesn't ;-)  You can put them back to `[[`

jviotti:
True, fixed!


—
Juan Cruz Viotti

--- original message ---
On Mon, Jan 30, 2017 at 01:43 pm, <notifications@github.com> Andrew Scheller wrote:

@lurch commented on this pull request.
In scripts/darwin.sh:
> echo "protected: True"
else
echo "protected: False"
fi

- if [[ "$device" == "/dev/disk0" ]] || \
+ if [ "$device" = "/dev/disk0" ] || \
[[ ( "$removable" == "No" ) || ("$removable" == "Fixed") ]] || \
[[ ( "$location" =~ "Internal" ) && ( "$removable" != "Yes" ) && ( "$removable" != "Removable" ) ]] || \
Not when it's operating in bash-mode it doesn't ;-) You can put them back to [[
—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or mute the thread.
--- end of original message ---

scripts/darwin.sh
jviotti: True, fixed!

PR #135, opened 2017-01-30
Improve linux.sh based on shellcheck tips

jviotti:
These were the addressed issues:

```
In scripts/linux.sh line 8:
  /sbin/blkid -s UUID -o value $1*
                               ^-- SC2086: Double quote to prevent globbing and word splitting.

In scripts/linux.sh line 15:
DISKS="`lsblk -d --output NAME | ignore_first_line`"
       ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/linux.sh line 25:
  diskinfo=($(lsblk -b -d $device --output SIZE,RO,RM,MODEL | ignore_first_line))
                          ^-- SC2086: Double quote to prevent globbing and word splitting.

In scripts/linux.sh line 29:
  description=${diskinfo[@]:3}
              ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.

In scripts/linux.sh line 30:
  mountpoints=`get_mountpoint $device`
              ^-- SC2006: Use $(..) instead of legacy `..`.
                              ^-- SC2086: Double quote to prevent globbing and word splitting.

In scripts/linux.sh line 35:
    for uuid in `get_uuids $device`; do
                ^-- SC2006: Use $(..) instead of legacy `..`.
                           ^-- SC2086: Double quote to prevent globbing and word splitting.

In scripts/linux.sh line 36:
      mountpoints=$mountpoints`get_mountpoint /dev/disk/by-uuid/$uuid`
                              ^-- SC2006: Use $(..) instead of legacy `..`.
                                                                ^-- SC2086: Double quote to prevent globbing and word splitting.

In scripts/linux.sh line 43:
    subdevice=`echo "$device" | cut -d'/' -f3`
              ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/linux.sh line 45:
      description=`cat "/sys/class/block/$subdevice/device/name"`
                  ^-- SC2006: Use $(..) instead of legacy `..`.

In scripts/linux.sh line 58:
    echo "$mountpoints" | while read mountpoint ; do
                                ^-- SC2162: read without -r will mangle backslashes.

In scripts/linux.sh line 71:
  eval "`udevadm info \
        ^-- SC2006: Use $(..) instead of legacy `..`.
                                                                                             ^-- SC2086: Double quote to prevent globbing and word splitting.
```

### DIFF
--- a/scripts/linux.sh
+++ b/scripts/linux.sh
@@ -1,18 +1,21 @@
 #!/bin/bash
 
-function ignore_first_line {
+set -u
+set -e
+
+ignore_first_line() {
   tail -n +2
 }
 
-function get_uuids {
-  /sbin/blkid -s UUID -o value $1*
+get_uuids() {
+  /sbin/blkid -s UUID -o value "$1"*
 }
 
-function get_mountpoint {
+get_mountpoint() {
   df --output=source,target | grep "^$1" | awk '!($1="")&&gsub(/^ /,"")'
 }
 
-DISKS="`lsblk -d --output NAME | ignore_first_line`"
+DISKS="$(lsblk -d --output NAME | ignore_first_line)"
 
 for disk in $DISKS; do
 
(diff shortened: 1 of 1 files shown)


jviotti:
BTW, we should put some effort into making this script has POSIX compliance, to avoid any weird issue on non-mainstream Linux distributions.

lurch:
There's a difference between POSIX (sh) compliance, and POSIX (bash) compliance.
And strictly-speaking POSIX is a subset of Linux, but it's fine to make this script only general-Linux-compliant, and not strict-POSIX-compliant.
(if it _was_ strict-POSIX-compliant then the same script should run on both Linux and OS X, but that's obviously not what we're aiming for here, as there's bound to be platform-specific parts)

scripts/linux.sh
```diff
@@ -22,27 +25,27 @@ for disk in $DISKS; do
...
+    for uuid in $(get_uuids "$device"); do
+      mountpoints="$mountpoints$(get_mountpoint "/dev/disk/by-uuid/$uuid")"
     done
   fi
 
   # If we couldn't get the description from `lsblk`, see if we can get it
   # from sysfs (e.g. PCI-connected SD cards that appear as `/dev/mmcblk0`)
   if [ -z "$description" ]; then
-    subdevice=`echo "$device" | cut -d'/' -f3`
+    subdevice="$(echo "$device" | cut -d '/' -f3)"
```
lurch: If you're adding a space after `-d`, should there also be a space after `-f` ? ;)

scripts/linux.sh
```diff
@@ -1,18 +1,21 @@
...
-function get_uuids {
-  /sbin/blkid -s UUID -o value $1*
+get_uuids() {
+  /sbin/blkid -s UUID -o value "$1"*
 }
 
-function get_mountpoint {
-  df --output=source,target | grep "^$1" | awk '!($1="")&&gsub(/^ /,"")'
+get_mountpoint() {
+  mount | grep "^$1" | cut -d ' ' -f3
```
lurch: What prompted this change from `df` to `mount`? IMHO it's not related to the shellcheck output, and so should probably be in a separate PR?

scripts/linux.sh
jviotti: Good catch!

scripts/linux.sh
jviotti: Oops, this was a bug fix I had in my branch that I never sent. I'll split to another PR.

jviotti: See https://github.com/resin-io-modules/drivelist/pull/136

PR #136, opened 2017-01-30
Consider multiple mountpoints on GNU/Linux

jviotti:
Turns out `df` only reports th *first* mountpoint. We switch to
`/proc/mounts` as the source of truth.

### DIFF
--- a/scripts/linux.sh
+++ b/scripts/linux.sh
@@ -11,8 +11,8 @@ get_uuids() {
   /sbin/blkid -s UUID -o value "$1"*
 }
 
-get_mountpoint() {
-  df --output=source,target | grep "^$1" | awk '!($1="")&&gsub(/^ /,"")'
+get_mountpoints() {
+  grep "^$1" /proc/mounts | cut -d ' ' -f 2
 }
 
 DISKS="$(lsblk -d --output NAME | ignore_first_line)"
@@ -30,13 +30,13 @@ for disk in $DISKS; do
   protected=${diskinfo[1]}
   removable=${diskinfo[2]}
   description=${diskinfo[*]:3}
-  mountpoints="$(get_mountpoint "$device")"
+  mountpoints="$(get_mountpoints "$device")"
 
   # If we couldn't get the mount points as `/dev/$disk`,
   # get the disk UUIDs, and check as `/dev/disk/by-uuid/$uuid`
   if [ -z "$mountpoints" ]; then
     for uuid in $(get_uuids "$device"); do
-      mountpoints="$mountpoints$(get_mountpoint "/dev/disk/by-uuid/$uuid")"
+      mountpoints="$mountpoints$(get_mountpoints "/dev/disk/by-uuid/$uuid")"
     done
   fi
 


lurch:
Hmmm? Can you give an example of where `mount` works, where `df` doesn't?
I wonder if reading `/proc/mounts` directly would be better than parsing the output of the `mount` command?

jviotti:
Sure. I have an USB drive on `/dev/sdc1`, which is being mounted at `/media/jviotti/U` automatically:

```sh
$ mount
...
/dev/sdc1 on /media/jviotti/U type vfat (rw,nosuid,nodev,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,showexec,utf8,flush,errors=remount-ro,uhelper=udisks2)
...
```

Now I can mount it somewhere else as well:

```sh
# mkdir /media/test
# mount /dev/sdc1 /media/test
```

The `mount` command shows both mount points:

```sh
$ mount
...
/dev/sdc1 on /media/jviotti/U type vfat (rw,nosuid,nodev,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,showexec,utf8,flush,errors=remount-ro,uhelper=udisks2)
/dev/sdc1 on /media/test type vfat (rw,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,showexec,utf8,flush,errors=remount-ro)
...
```

However `df` shows:

```sh
$ df --output=source,target | grep "^/dev/sdc"
/dev/sdc1      /media/test
```

Which is only the last mount point.

> I wonder if reading /proc/mounts directly would be better than parsing the output of the mount command?

Yeah, that's probably right. I'll update the PR in a minute.

scripts/linux.sh
```diff
@@ -12,7 +12,7 @@ get_uuids() {
 }
 
 get_mountpoint() {
```
lurch: Maybe this should be renamed to `get_mountpoints` to make it clearer that it _might_ return a (multi-line) list?

scripts/linux.sh
```diff
@@ -12,7 +12,7 @@ get_uuids() {
...
-  df --output=source,target | grep "^$1" | awk '!($1="")&&gsub(/^ /,"")'
+  cat /proc/mounts | grep "^$1" | cut -d ' ' -f 2
```
lurch: This can be simplified to
`grep "^$1" /proc/mounts | cut -d ' ' -f 2`

jviotti: Done!

scripts/linux.sh
jviotti: Done!

lurch:
Looks like `/proc/mounts` might be more accurate too?
```bash
$ mount | grep sda
/dev/sda4 on / type ext4 (rw,errors=remount-ro)
/dev/sda1 on /boot/efi type vfat (rw)
$ grep sda /proc/mounts
/dev/sda4 / ext4 rw,relatime,errors=remount-ro,data=ordered 0 0
/dev/sda1 /boot/efi vfat rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,errors=remount-ro 0 0
/dev/sda4 /var/lib/docker/aufs ext4 rw,relatime,errors=remount-ro,data=ordered 0 0
```

jviotti:
Cool. Weird that `mount` doesn't detect that.

PR #137, opened 2017-01-30
Print stderr script output to stderr if exit code is non zero

jviotti:
Platform scripts might output warnings to `stderr` that we don't want to
take into consideration if the script exits with an exit code 0.

If the scripts exits with a non-zero code, `.execFile()` yields an error
whose message includes the `stderr` output, so we don't need to do
anything ourselves.

See: https://github.com/resin-io/etcher/issues/1034

### DIFF
--- a/lib/scripts.js
+++ b/lib/scripts.js
@@ -18,6 +18,7 @@
 
 const childProcess = require('child_process');
 const path = require('path');
+const debug = require('debug')(require('../package.json').name);
 
 /**
  * @summary Absolute path to platform scripts
@@ -86,8 +87,12 @@ exports.run = (script, callback) => {
       return callback(error);
     }
 
+    // Don't throw an error if we get `stderr` output from
+    // the drive detection scripts at this point, given that
+    // if the script already exitted with code zero, then
+    // we consider them warnings that we can safely ignore.
     if (stderr.trim().length) {
-      return callback(new Error(stderr));
+      debug('stderr: %s', stderr);
     }
 
     return callback(null, stdout);
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "mochainon": "^1.0.0"
   },
   "dependencies": {
+    "debug": "^2.6.0",
     "js-yaml": "^3.4.1",
     "lodash": "^4.16.4"
   }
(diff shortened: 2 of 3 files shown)


jviotti:
Oops, I forgot to update the test suite.

lib/scripts.js
```diff
@@ -82,12 +83,21 @@ exports.paths = {
  */
 exports.run = (script, callback) => {
   childProcess.execFile(script, (error, stdout, stderr) => {
+
+    // Platform scripts might output warnings to `stderr`
+    // that we don't want to take into consideration
+    // if the script exits with an exit code 0.
```
lurch: IMHO this comment is confusing - any chance you could reword it? Also, perhaps it would make more sense if this comment was moved to just above the `if (stderr.trim().length) {` line?

lib/scripts.js
```diff
@@ -82,12 +83,21 @@ exports.paths = {
...
+    // yields an error whose message includes the `stderr`
+    // output, so we don't need to do anything ourselves.
+
     if (error) {
       return callback(error);
     }
 
     if (stderr.trim().length) {
-      return callback(new Error(stderr));
+      debug('stderr: %s', stderr);
```
lurch: And could you add a unit-test for this `debug` too?

lib/scripts.js
jviotti: What about now?

lib/scripts.js
jviotti: Might be tricky, but let me see what I can do.

jviotti: Yeah, this might require a lot of work. `debug()` is simply a wrapper around `console.error`, which we could stub, however `debug()` checks if the DEBUG environment variable is set when this file is first interpreted (therefore we can't modify it at runtime), so I guess testing this would be overkill.

lib/scripts.js
lurch: Yup, much better :-)

lib/scripts.js
lurch: I don't know much about stubbing and nodejs unit tests, but would it be possible to stub `debug()` instead?

jviotti: Sadly not, because its a standalone function. This would have been possible if the function returns an object containing a `.log()` function for instance.

PR #141, opened 2017-02-15
Omit empty internal SD Card readers on Windows

jviotti:
Windows lists internal SD Card readers by default, even when they have
no SD Card inserted on them. As a solution, we can check if the size
returned for such internal reader is `null`, and if so, omit it from the
final drive list.


WasabiFan:
This isn't just for internal readers; in my case, it was a USB multi-card reader.

PR #144, opened 2017-02-20
Implement drivelist.get()

jviotti:
This function allows the user to get information about a single drive.
Currently, this function runs the whole drive detection script and
filters the output, however in the future we might update the scripts to
take a drive argument, and only print information about that one, which
should be way faster.

See: https://github.com/resin-io/etcher/issues/750

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,18 +191,23 @@ $ npm install --save drivelist
 Documentation
 -------------
 
+
+* [drivelist](#module_drivelist)
+    * [.list(callback)](#module_drivelist.list)
+    * [.get(device, callback)](#module_drivelist.get)
+
 <a name="module_drivelist.list"></a>
 
 ### drivelist.list(callback)
-**Kind**: static method of <code>[drivelist](#module_drivelist)</code>
-**Summary**: List available drives
-**Access:** public
+**Kind**: static method of <code>[drivelist](#module_drivelist)</code>  
+**Summary**: List available drives  
+**Access:** public  
 
 | Param | Type | Description |
 | --- | --- | --- |
 | callback | <code>function</code> | callback (error, drives) |
 
-**Example**
+**Example**  
 ```js
 const drivelist = require('drivelist');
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -191,18 +191,23 @@ $ npm install --save drivelist
...
+
 <a name="module_drivelist.list"></a>
 
 ### drivelist.list(callback)
-**Kind**: static method of <code>[drivelist](#module_drivelist)</code>
-**Summary**: List available drives
-**Access:** public
+**Kind**: static method of <code>[drivelist](#module_drivelist)</code>  
+**Summary**: List available drives  
+**Access:** public
```
lurch: What's with the extra spaces?

README.md
```diff
@@ -249,5 +278,3 @@ License
 -------
 
 The project is licensed under the Apache 2.0 license.
-
-[yaml]: http://yaml.org
```
lurch: I think this removal was a mistake? (the `[yaml]` link is still in use)

jviotti: Oops, fixed!

README.md
jviotti: It seems to come from the JSDoc to markdown module. I could manually edit it, but then they will re-appear the next time we re-generate it.

lurch: Ahhh, fair enough!

tests/drivelist.spec.js
```diff
@@ -166,4 +166,105 @@ describe('Drivelist', function() {
...
 
+  describe('.get()', function() {
+
+    describe('given scripts run succesfully', function() {
+
+      beforeEach(function() {
+        this.scriptsRunStub = m.sinon.stub(scripts, 'run');
+
+        this.scriptsRunStub.yields(null, [
+          'device: "\\\\\\\\.\\\\PHYSICALDRIVE1"',
```
lurch: slash-tastic ;-)

README.md
jhermsmeier: The two spaces force a line break, so that those don't get squished into the same line.

> When you do want to insert a \<br /\> break tag using Markdown, you end a line with two or more spaces, then type return

http://daringfireball.net/projects/markdown/syntax#p

jviotti:
Lets implement this on the native modules instead.

lurch:
Is it worth keeping this open just as a reminder?

jviotti:
Lets create an issue as a reminder: https://github.com/resin-io-modules/drivelist/issues/163

PR #145, opened 2017-02-21
Fix missing executable file mode on scripts

jhermsmeier:
As it turns out, `npm` won't mark files as executable post install, unless they're listed in the package.json's `bin` key.


jviotti:
Amazing stuff! There is no issue to link to. I'll merge once you update the commit message to remove the "Connects to" tag!

jhermsmeier:
> remove the "Connects to" tag!

It's only in the PR description, not the commit message (but also removed it from there now) ;)

jviotti:
Ah, sorry about that! :)

Issue #147, opened 2017-03-03
diskinfo[0]: unbound variable


Edudjr:
I was trying to get a list of devices every half a second, like the code below:  
```javascript
setInterval(function(){
  drivelist.list((error, drives) => {
    if (error) {
      throw error;
    }
    callback(drives);
  });
}, 500);
```

However, I was getting an error of unbound variable whenever I unplugged the usb flash drive: 
![Unbound variable](http://i.imgur.com/0adPmPl.png)

It seems that this modification to the line 29 solved my problem:
```bash
#Check for diskinfo before continuing
if [ -z "${diskinfo-}" ]; then
  continue
fi
size=${diskinfo[0]}
...
```
Could you check that?

jviotti:
Thanks a lot for reporting this and investigating and potential solution. We're hitting this same issue on Etcher (https://github.com/resin-io/etcher/issues/1126#issuecomment-282083506), which is great. I'll push a fix today.

Edudjr:
@jviotti great! I hope this will fix the issue :)

jviotti:
See https://github.com/resin-io-modules/drivelist/pull/148. I fixed for Linux and OS X, and working on the Windows one, which proved to be trickier.

jviotti:
@Edudjr I finally got the Windows VBScript working \o/ Feel free to review the PR. I'll publish a new version to npm once its merged. Thanks a lot!

Edudjr:
@jviotti Awesome! Thanks for reviewing this issue so quickly :)

PR #148, opened 2017-03-03
Prevent errors when unplugging drives while the scripts are running

jviotti:
macOS

- Check the `diskutil info` exit code

GNU/Linux

- Check the `diskinfo` list (thanks @Edudjr!)

Windows

- Create `IsWMIObjectValid` utility
- Make `GetLogicalDisks` return `Nothing` if the WMI query failed
- Check for `Nothing` in the main loop, and ignore the drive if that's
  the case

Fixes: https://github.com/resin-io-modules/drivelist/issues/147
See: https://github.com/resin-io/etcher/issues/1126

### DIFF
--- a/scripts/linux.sh
+++ b/scripts/linux.sh
@@ -26,6 +26,13 @@ for disk in $DISKS; do
 
   device="/dev/$disk"
   diskinfo=($(lsblk -b -d "$device" --output SIZE,RO,RM,MODEL | ignore_first_line))
+
+  # Omit drives for which `lsblk` failed, which means they
+  # were unplugged right after we got the list of all drives
+  if [ -z "${diskinfo-}" ]; then
+    continue
+  fi
+
   size=${diskinfo[0]}
   protected=${diskinfo[1]}
   removable=${diskinfo[2]}
--- a/scripts/win32.bat
+++ b/scripts/win32.bat
@@ -69,78 +69,93 @@ Function GetOperatingSystemDevice()
 	Next
 End Function
 
+Function IsWMIObjectValid(WMIObject)
+	On Error Resume Next
+	Columns = WMIObject.Count
+	IsWMIObjectValid = True
+	If (Err.Number <> 0) Then
+		IsWMIobjectValid = False
+	End If
+	On Error Goto 0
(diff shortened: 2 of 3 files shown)


lurch:
IMHO this PR should be regarded as WIP until the Windows script also gets updated similarly (we should aim to keep consistent behaviour across all three platforms).

jviotti:
@lurch My bad! The Windows part is done!

jviotti:
I couldn't avoid the extra indentation in some parts of the Windows script, given that VBScript has nothing similar to the `continue` keyword to go straight to the next loop iteration. The alternative was to create global Boolean variables, so I guess a bit of indentation is better :)

scripts/win32.bat
```diff
@@ -69,78 +69,93 @@ Function GetOperatingSystemDevice()
 	Next
 End Function
 
+Function IsWMIObjectValid(WMIObject)
+	On Error Resume Next
```
lurch: Why is this only a one-space indent, when the rest of the file seems to use two-spaces?

lurch:
> I couldn't avoid the extra indentation in some parts of the Windows script,

That's fine, it's just a shame the `diff` displayed by github isn't more tolerant to changing indent-levels ;-)

scripts/linux.sh
```diff
@@ -26,6 +26,13 @@ for disk in $DISKS; do
 
   device="/dev/$disk"
   diskinfo=($(lsblk -b -d "$device" --output SIZE,RO,RM,MODEL | ignore_first_line))
```
jviotti: It already get's silently swallowed :)

scripts/win32.bat
jviotti: Hm, maybe some GitHub weirdness? I see tabs locally:

screenshot 2017-03-03 14 43 29

jviotti:
@lurch Do I have the green sign to merge? :)

scripts/win32.bat
lurch: Ahh, sorry, didn't realise it was just a github-diff-display issue! ;-)

PR #151, opened 2017-03-06
Reset ERRORLEVEL at the beginning of win32.bat

jviotti:
Turns out ERRORLEVEL might carry errors from previous commands, given
that not every command resets it when exitting successfully.

This prevents us from getting false error alerts and helps us know for
sure that an exit code originated from drivelist.

See: https://github.com/resin-io/etcher/issues/1054

### DIFF
--- a/scripts/win32.bat
+++ b/scripts/win32.bat
@@ -1,6 +1,9 @@
 <!-- : Begin batch script
 @echo off
 
+:: Reset ERRORLEVEL. See http://stackoverflow.com/a/1113735/1641422
+ver > nul
+
 :: Ensure System32 is in the PATH, to avoid weird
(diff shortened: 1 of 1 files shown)


lurch:
Just out of curiosity - what runs before `win32.bat` that might be setting ERRORLEVEL to a non-zero status? Does that hint that there might be some other error somewhere that we're failing to catch?
(I'm not a DOS-scripting expert, but I worry that explicitly setting ERRORLEVEL to zero might be like a small child sticking their fingers in their ears? :wink: )

jviotti:
lol, turns out I was wrong. ERRORLEVEL is local to each session/script, so this PR is unnecessary. I'll close it!

Issue #152, opened 2017-03-15
Linux mount path contains invalid characters


ninox92:
I have a USB named `ECG REC`

When listing drives in Linux this results in a mountpoint path of:
/media/mypath/ECG\\\040REC

lurch:
Thanks for the report. Could you confirm that the updated version in #153 works properly now? :-)

jviotti:
I've published @lurch 's fix as v5.0.17

ninox92:
Fix comfirmed, can close!
Thanks for the quick fix <3

PR #153, opened 2017-03-15
Fix linux.sh to work when mountpoint contains spaces

lurch:
Connects to #152

Change-type: patch

### DIFF
--- a/scripts/linux.sh
+++ b/scripts/linux.sh
@@ -12,7 +12,7 @@ get_uuids() {
 }
 
 get_mountpoints() {
-  grep "^$1" /proc/mounts | cut -d ' ' -f 2
+  grep "^$1" /proc/mounts | cut -d ' ' -f 2 | sed 's,\\040, ,g'
 }
 
 DISKS="$(lsblk -d --output NAME | ignore_first_line)"


scripts/linux.sh
```diff
@@ -12,7 +12,7 @@ get_uuids() {
 }
 
 get_mountpoints() {
-  grep "^$1" /proc/mounts | cut -d ' ' -f 2
+  grep "^$1" /proc/mounts | cut -d ' ' -f 2 | sed 's,\\040, ,g'
```
jhermsmeier: Isn't `0x40` an `@`? Or how is this escape sequence to be interpreted?

lurch: It's not `0x40`, it's `040` in octal, i.e. `0x20` in hex or `32` in decimal, or `' '` in ASCII ;-)
This is because `/proc/mounts` is space-separated, and so it needs to use some escape-sequence when a mountpoint itself contains a space character.

jhermsmeier: Oh, right – don't mind me then, my brain didn't make the jump back from 32 when I read it as octal :)

lurch:
While looking at the `mountutils` code I stumbled across https://linux.die.net/man/3/setmntent which says there's actually 4 octal escape-codes in use: `space (\040), tab (\011), newline (\012) and backslash (\134)`

I guess we could add those other codes too for the sake of completion, or we could decide that those characters are _very_ unlikely to appear in a drive name / label / mountpoint.

Happy to submit another PR if you decide on the completion-approach @jviotti

jviotti:
Yeah, lets add them!

PR #156, opened 2017-03-23
Fix linux.sh to work with mountpoints containing tab, newline or backslash

lurch:
I tested mountpoints on Linux containing all kinds of weird characters, and `drivelist` now works properly with all of them. This is what running `node example.js` outputs:
```json
[
  {
    "device": "/dev/sda",
    "description": "SanDisk SD8SN8U1",
    "size": 1024209543168,
    "mountpoints": [
      {
        "path": "/"
      },
      {
        "path": "/boot/efi"
      },
      {
        "path": "/home"
      },
      {
        "path": "/var/lib/docker/aufs"
      }
    ],
    "raw": "/dev/sda",
    "protected": false,
    "system": true
  },
  {
    "device": "/dev/sdb",
    "description": "Cruzer",
    "size": 2018508288,
    "mountpoints": [
      {
        "path": "/tmp/dodgy/amp&dir"
      },
      {
        "path": "/tmp/dodgy/bs\\dir"
      },
      {
        "path": "/tmp/dodgy/dollar$dir"
      },
      {
        "path": "/tmp/dodgy/dq\"dir"
      },
      {
        "path": "/tmp/dodgy/hash#dir"
      },
      {
        "path": "/tmp/dodgy/hat^dir"
      },
      {
        "path": "/tmp/dodgy/nl\ndir"
      },
      {
        "path": "/tmp/dodgy/space dir"
      },
      {
        "path": "/tmp/dodgy/sq'dir"
      },
      {
        "path": "/tmp/dodgy/star*dir"
      },
      {
        "path": "/tmp/dodgy/tab\tdir"
      }
    ],
    "raw": "/dev/sdb",
    "protected": false,
    "system": false
  }
]
```

@jviotti I leave it up to you to make sure that `darwin.sh` works equally as well ;-)

Just for reference, here's the outputs of `scripts/linux.sh`:
```
device: /dev/sda
description: "SanDisk SD8SN8U1"
size: 1024209543168
mountpoints:
  - path: "/"
  - path: "/boot/efi"
  - path: "/home"
  - path: "/var/lib/docker/aufs"
raw: /dev/sda
protected: False
system: True

device: /dev/sdb
description: "Cruzer"
size: 2018508288
mountpoints:
  - path: "/tmp/dodgy/amp&dir"
  - path: "/tmp/dodgy/bs\\dir"
  - path: "/tmp/dodgy/dollar$dir"
  - path: "/tmp/dodgy/dq"dir"
[...]

PR #161, opened 2017-04-18
Correctly deal with unicode characters

jviotti:
This seems to be the cause of many Etcher bugs. This commit deals with
the following issues:

- Unicode characters
- Non printable ascii characters

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -65,23 +65,44 @@ module.exports = (input) => {
   }
 
   return _.compact(_.map(input.split(/\n\s*\n/), (device) => {
-
     device = _.chain(device)
       .split('\n')
       .filter((line) => {
         return /^(\s\s-\s)?[a-z]+:/g.test(line);
       })
       .map((line) => {
-        return line.replace(/"/g, (match, index, string) => {
-          if (_.some([
-            string.indexOf('"') === index,
-            string.lastIndexOf('"') === index
-          ])) {
-            return match;
-          }
+        return line
+          .replace(/\\[^.\\]/g, (match, index, string) => {
+            const escapedCharacter = _.last(match);
 
-          return '\\"';
-        });
+            if (string[index - 1] === '\\' || _.includes([
+              'b',
+              'f',
+              'n',
+              'r',
+              't',
+              'v'
+            ], escapedCharacter)) {
+              return match;
+            }
+
+            return escapedCharacter;
+          })
+
+          // Remove non printable ascii characters
+          // See http://stackoverflow.com/a/24229554
+          .replace(/[^\x20-\x7E]+/g, '')
+
+          .replace(/"/g, (match, index, string) => {
+            if (_.some([
+              string.indexOf('"') === index,
+              string.lastIndexOf('"') === index
+            ])) {
+              return match;
+            }
+
+            return '\\"';
+          });
       })
       .join('\n')
       .value();
--- a/tests/parse.spec.js
+++ b/tests/parse.spec.js
@@ -232,4 +232,45 @@ describe('Parse', function() {
     ]);
   });
 
+  it('should remove the backslash of unknown escape sequences inside values', function() {
+    m.chai.expect(parse([
+      'device: /dev/disk1',
+      'description: "\\]=-01`23456\\78\\90\\=\\-0\\98e"'
+    ].join('\n'))).to.deep.equal([
+      {
+        device: '/dev/disk1',
+        description: ']=-01`234567890=-098e'
+      }
+    ]);
+  });
+
+  it('should maintain known escape sequences', function() {
+    m.chai.expect(parse([
+      'device: /dev/disk1',
+      'description: "foo\\bbar\\f\\nbaz\\r\\tqux\\v"'
+    ].join('\n'))).to.deep.equal([
+      {
+        device: '/dev/disk1',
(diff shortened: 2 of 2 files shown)


tests/parse.spec.js
```diff
@@ -232,4 +232,54 @@ describe('Parse', function() {
...
+      'description: "foo\bbar\rbaz"'
+    ].join('\n'))).to.deep.equal([
+      {
+        device: '/dev/disk1',
+        description: 'foobarbaz'
+      }
+    ]);
+  });
+
+  it('should not remove tabs inside values', function() {
```
lurch: Huh? Seems like it's actually replacing tabs by 4 spaces. If that _is_ the intended behaviour, this test should probably be renamed ;-)

tests/parse.spec.js
```diff
@@ -232,4 +232,54 @@ describe('Parse', function() {
...
+      'description: "\\]=-01`234567890=-098e"'
+    ].join('\n'))).to.deep.equal([
+      {
+        device: '/dev/disk1',
+        description: ']=-01`234567890=-098e'
+      }
+    ]);
+  });
+
+  it('should delete known escape sequences', function() {
```
lurch: Does this impact on the changes I made in #156 ?

lurch: (and if TAB is a special-case, maybe this test should be called `it('should delete known escape sequences except TAB', function() {` ?)

lib/parse.js
```diff
@@ -65,23 +65,36 @@ module.exports = (input) => {
...
+            if (string[index - 1] === '\\') {
+              return match;
+            }
+
+            return _.last(match);
+          })
+
+          // Remove non printable ascii characters
+          // See http://stackoverflow.com/a/24229554
+          .replace(/[^\x20-\x7E]+/g, '')
```
lurch: Heh, very similar to the regex at https://github.com/resin-io/etcher/blob/master/scripts/ci/ensure-all-text-files-only-ascii.sh#L31 ;-)

tests/parse.spec.js
```diff
@@ -232,4 +232,54 @@ describe('Parse', function() {
     ]);
   });
 
+  it('should handle unknown escape sequences inside a value', function() {
```
lurch: IMHO we should also have a test on a string containing multiple unknown escape sequences

lurch: ...and what exactly does "handle" mean here?

tests/parse.spec.js
```diff
@@ -232,4 +232,54 @@ describe('Parse', function() {
...
+      'description: "foo\tbar"'
+    ].join('\n'))).to.deep.equal([
+      {
+        device: '/dev/disk1',
+        description: 'foo    bar'
+      }
+    ]);
+  });
+
+  it('should handle unicode characters inside a value', function() {
```
lurch: Should this also be testing printable unicode characters in addition to unprintable unicode characters?
And does "should handle" here actually mean "should delete"?

lurch:
Do we have any idea where these unicode characters and unknown escape sequences are actually coming from?

tests/parse.spec.js
jviotti: I spammed the string with more backslashes, and clarified what "handle" means!

tests/parse.spec.js
jviotti: I think the changes are complementary. Your changes make the linux script converts things like `\011` into ASCII escape sequences that can be removed here, but do you think we can move that sed logic here, so we make sure it gets applied everywhere?

> (and if TAB is a special-case, maybe this test should be called it('should delete known escape sequences except TAB', function() { ?)

Fixed!

tests/parse.spec.js
jviotti: Yeah, good catch. Fixed!

tests/parse.spec.js
jviotti: Makes sense. I added some printable characters, and clarified the message.

jviotti:
@lurch All the drive descriptions I added to the test suite are legit descriptions I obtained from Sentry. Interestingly enough, the 92% of these occurrences come from GNU/Linux systems, but I have no clue how they got there in the first place.

tests/parse.spec.js
lurch: Yeah, that's a good point. My changes in #156 convert the Linux mount-specific escape sequences to 'regular ASCII escapes', and then your changes here remove those regular ASCII escape sequences. That sounds like the correct separation of logic.

Although now I think about it, perhaps the changes you're making in this PR are Etcher-specific, and there might be other users of `drivelist` who _want_ these escape sequences to be preserved? Perhaps there should be an optional parameter that a user could supply to drivelist, specifying whether they want these escape sequences preserved or not?

lurch:
Hmmm, interesting. I wonder if any of them might be caused by data-corruption, and perhaps Linux is more tolerant at 'preserving' corrupted data and passing it back to the user? Or maybe it's Windows or OSX specific codepage issues which Linux isn't able to properly handle? Who knows... ;-) \*shrug\*

tests/parse.spec.js
jviotti: > Perhaps there should be an optional parameter that a user could supply to drivelist, specifying whether they want these escape sequences preserved or not?

I think it makes a lot of sense. I made the escape sequences go through without issues, and we can handle from the Etcher side if really necessary.

PR #164, opened 2017-05-10
Inline platform scripts

jviotti:
This commit compiles the platform scripts at `scripts/` into a JSON file
that gets distributed with the code. The module then creates a temporary
file, writes the content of the JSON file, and executes the file.

The reason behind this change are:

- When running on an Electron `asar` archive, the script will be
  automatically extracted to a temporary location. If this temporary
  location contains an ampersand (e.g. when the Windows username
  contains an ampersand), then the execution will fail. We tried every
  possible quoting workaround without success. The only thing that
  worked was setting the current working directory to the directory name
  of the temporary file, and execute the basename, however we can't do
  that without having control over the extraction mechanism

- Since the platform scripts live as separate non-JavaScript files,
  concatenators like Browserify or WebPack fail to include them on the
  final package

See: https://github.com/resin-io/etcher/issues/1395


tests/execute.spec.js
```diff
@@ -59,7 +38,10 @@ describe('Scripts', function() {
       });
 
       it('should yield the error', function(done) {
-        scripts.run('foo', (error, output) => {
+        execute.extractAndRun({
+          content: '#!/bin/bash\necho "Hello"',
```
lurch: Given the way that things are being stubbed, I think you should make it clear that the `content` here isn't actually being used, e.g. by changing it to `content: 'dummy content'`.

And given this new extractAndRun mechanism, I think you should add some platform-specific tests that *do* actually test it. (e.g. `if (os.platform === 'win32') { extractAndRun('@echo off\n@echo Hello world\n') } else { '#!/bin/bash\necho "Hello World"\n' }`

And would it be worth having a unit test that the scripts embedded inside `scripts.json` actually do match the platform scripts in the repo?

lurch:
Looks good, but don't the "encoded scripts" actually need to live inside the JS file that gets executed, rather than in some external JSON file? (if I've understood the way that Electron asars work)

tests/execute.spec.js
jviotti: Sounds good. I'll address all these testing comments in a minute.

jviotti:
The contents are stored in a JSON file, which can required with `require()`, which is already patched by Electron, so anything that is either JS or JSON will do the trick.

lurch:
Oh okay, I didn't realise that Electron patched `require` too ;P

.eslintrc.yml
```diff
@@ -212,7 +212,7 @@ rules:
   no-label-var:
     - error
   no-shadow:
-    - error
+    - off
```
lurch: Just out of sheer curiosity - what does this change mean? Would it be better to add exclusions in the one or two places that need it?

doc/README.hbs
```diff
@@ -17,6 +17,53 @@ Supports:
 - GNU/Linux distributions that include [util-linux](https://github.com/karelzak/util-linux) and [udev](https://wiki.archlinux.org/index.php/udev).
 - Mac OS X.
 
+The `drivelist` core consists of a set of scripts built with technologies that
```
lurch: Maybe worth adding an inline comment at the top of this file, telling people to edit `doc/README.hbs` rather than `README.md` directly? ;)

lib/execute.js
```diff
@@ -0,0 +1,91 @@
...
+
+    fs.writeFileAsync(fileDescriptor, script.content).return(fileDescriptor).then(fs.closeAsync).then(() => {
+      return new Bluebird((resolve, reject) => {
+
+        // On Windows, if the directory containing the temporary path
+        // includes an ampersand (e.g. if the username contains an ampersand)
+        // then the execution will fail.
+        // Turns out we can pass the temporary script dirname as the current
+        // working directory and execute the basename, and then it all works.
+        return childProcess.execFile(path.basename(temporaryPath), {
```
lurch: Now we're using this new self-extracting script mechanism, can't we just change this to `.exec` instead of `.execFile`, which _does_ handle ampersands correctly?

lib/scripts.json
```diff
@@ -0,0 +1,14 @@
...
+    "filename": "darwin.sh"
```
lurch: If we're extracting to a temporary path anyway, does it even matter what filename we use? I guess technically it's only on Windows where the file-extension matters.

lib/execute.js
```diff
@@ -0,0 +1,91 @@
...
+const debug = require('debug')(require('../package.json').name);
+const Bluebird = require('bluebird');
+const fs = Bluebird.promisifyAll(require('fs'));
+const childProcess = require('child_process');
+const path = require('path');
+const tmp = require('tmp');
+tmp.setGracefulCleanup();
+
+/**
+ * @summary Run a platform script
```
lurch: This whole docstring obviously needs updating ;)

tests/execute.spec.js
jviotti: Done!

lib/execute.js
jviotti: Done!

lib/scripts.json
jviotti: Sure, but I didn't want to hard-code `.bat` on the execute function, so I thought it'd be nicer to preserve the original filename, and get the extension from it.

doc/README.hbs
jviotti: Haha, done!

.eslintrc.yml
jviotti: I wanted to shut up the linter while I was hacking on it. I'll fix it now.

doc/README.hbs
```diff
@@ -1,3 +1,5 @@
+<!-- Make sure you edit doc/README.hbs rather than README.md -->
```
lurch: Personally, I'd add 'because the latter is auto-generated' at the end ;)

lib/execute.js
jviotti: Interesting. I'll give it a go, and I'll update the code if it works fine.

jviotti: You're right, `.exec` works fine. Awesome discovery. I'll update the PR :)

lib/drivelist.js
```diff
@@ -46,14 +47,14 @@ const scripts = require('./scripts');
  */
 exports.list = (callback) => {
   const operatingSystem = os.platform();
-  const script = scripts.paths[operatingSystem];
+  const script = scripts[operatingSystem];
```
lurch: minor nit, but if we renamed this from `script` to `script_text` then that might make these changes more obvious? ;)

lib/execute.js
jviotti: BTW, this does look like a Node.js bug. It will be fun to look into it later :)

jviotti:
All the comments have been addressed!

lib/execute.js
```diff
@@ -0,0 +1,119 @@
...
+const debug = require('debug')(require('../package.json').name);
+const Bluebird = require('bluebird');
+const fs = Bluebird.promisifyAll(require('fs'));
+const childProcess = require('child_process');
+const path = require('path');
+const tmp = require('tmp');
+tmp.setGracefulCleanup();
+
+/**
+ * @summary Create a temporary script file
```
lurch: Can we add "with the given contents" here?

lib/execute.js
```diff
@@ -0,0 +1,119 @@
...
+ * });
+ */
+exports.extractAndRun = (script, callback) => {
+  const extension = path.extname(script.originalFilename);
+  createTemporaryScriptFile(extension, script.content).then((temporaryPath) => {
+    return new Bluebird((resolve, reject) => {
+
+      // On Windows, if the directory containing the temporary path
+      // includes an ampersand (e.g. if the username contains an ampersand)
+      // then `.execFile` will fail, even with proper quoting.
```
lurch: This comment could be removed now?
EDIT: Or you could add "which is why we're using `.exec` instead" on the end ;)

tests/execute.spec.js
```diff
@@ -0,0 +1,158 @@
...
+        return {
+          content: '#!/bin/bash\necho "foo bar baz"',
+          originalFilename: 'hello.sh'
+        };
+      });
+
+      execute.extractAndRun(script, (error, output) => {
+        m.chai.expect(error).to.not.exist;
+
+        if (os.platform() === 'win32') {
```
lurch: ...or I guess we could trim the output, getting rid of the need for this platform-detection?

tests/scripts.spec.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Resin.io
+ * Copyright 2017 Resin.io
```
lurch: Yeah, that's fair given how much this file has changed :)

doc/README.hbs
jviotti: Done!

lib/execute.js
jviotti: Done!

lib/scripts.json
```diff
@@ -0,0 +1,14 @@
...
+    "originalFilename": "darwin.sh"
```
lurch: Hmmm, after leaving my approval comment, it's occurred to me that it might be useful to add a `"type": "text"` attribute to each file, to give us the flexibility to later add base64-encoded binary files too? (or maybe we should leave it as-is and only cross that bridge when we get to it)

lib/execute.js
jviotti: I'll add a clarification to the end of the comment!

tests/execute.spec.js
jviotti: Sounds good to me!

lib/scripts.json
jviotti: Makes a lot of sense. Added!

lib/drivelist.js
jviotti: Keep in mind that `script` is an object at this point.

lurch: Oh right, even more reason to give it a variable name that better reflects what it is, and that it's not just a filename the way that it used to be ;)

tests/execute.spec.js
lurch: (as long as you add a comment saying that the trim is only to get rid of platform-specific line-endings)

lib/drivelist.js
jviotti: Done!

README.md
```diff
@@ -240,6 +242,12 @@ Before submitting a PR, please make sure that you include tests, and that the li
 $ npm run lint
 ```
 
+Execute the following command after making any changes to the platform scripts:
+
+```sh
+node scripts/compile.js
```
lurch: Should we add a 'npm run' script to `package.json` for that? :)

jviotti: Done!

lib/execute.js
```diff
@@ -0,0 +1,120 @@
...
+};
+
+/**
+ * @summary Extract and run an inline script
+ * @function
+ * @public
+ *
+ * @param {Object} script - script
+ * @param {String} script.content - script content
+ * @param {String} script.originalFilename - script original filename
```
lurch: Should we document script.type here too?

lib/execute.js
```diff
@@ -0,0 +1,120 @@
...
+ * }, (error, output) => {
+ *   if (error) {
+ *     throw error;
+ *   }
+ *
+ *   console.log(output);
+ * });
+ */
+exports.extractAndRun = (script, callback) => {
+  const extension = path.extname(script.originalFilename);
```
lurch: ...and then in here we could do:
```
if (script.type === "text") {
    createTemporaryScriptFile(....
    etc.
} else {
   throw Error(`Don't know how to extract ${type}`);
}
```
?

jviotti: Can we leave it like this until we do get a use case for it? (e.g: for now assume that everything is type text)

lurch: Yeah sure - that's why I said "(or maybe we should leave it as-is and only cross that bridge when we get to it)" but then you added it anyway ;)

lurch:
So... a crazy thought that popped into my head over the weekend - why store `lib/scripts.json` in the repo at all - couldn't it just be generated automatically when you `npm install` this module, and then there'd be no need to remember to manually update it?

jviotti:
Sounds good to me, we just have to be extra careful that there is no possible state where everything will appear to have succeeded but the json file is not generated, since that will get us into some nasty issues

lurch:
Created #165 :)

PR #166, opened 2017-05-26
Use both `volume_name` and `description` when applicable

dustMason:
See discussion on #89


jviotti:
Awesome, that was fast :)

jviotti:
I published 5.0.21 with the fix!

dustMason:
Fantastic, thanks for the quick turnaround!

PR #168, opened 2017-06-06
Don't use perl when fetching the drive size on macOS

jviotti:
For some strange reason, this particular perl line drives the CPU to
100% on some cases, and remains alive even after drivelist is not
running anymore.

Fixes: https://github.com/resin-io/etcher/issues/1288


lurch:
Minor nit, but could you change "Get rid of perl" to "Don't use perl"? :)

jviotti:
Done!

PR #174, opened 2017-06-08
Adapt changelog for versionbot

lekkas:
Change-Type: patch
Fixes: #173


CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
...
----------
+# Change Log
 
-All notable changes to this project will be documented in this file.
+All notable changes to this project will be documented in this file
+automatically by Versionist. DO NOT EDIT THIS FILE MANUALLY!
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-### v5.0.22
+## v5.0.22
```
jviotti: Should you do this change for the other entries as well?

lekkas: I'll happily do for consistency reasons, let check with @hedss as well if this is a Versionbot requirement as well

hedss: It's down to `versionist` and I don't think it cares as long as it finds the latest and the tag.

lekkas: Fixes other version headers as well

PR #176, opened 2017-06-14
upgrade(package): Bump dependencies

jhermsmeier:
- eslint 3.9.0 -> 4.0.0
- jsdoc-to-markdown 2.0.1 -> 3.0.0
- mocha 2.4.5 -> 3.4.2

Change-Type: patch


lurch:
Just out of curiosity: Is this updating packages just to keep "up to date", or does this also get us fixes that we want / need??

lekkas:
just a heads up @jhermsmeier (or whoever merges this PR) you can apply the `procbots/versionbot/ready-to-merge` label to this PR instead of hitting the merge button. The versionbot will automatically add the changelog entries and cut a new version - note that npm publishing will still have to be done manually (at least for now)

jhermsmeier:
@lurch Both – naturally, if there's a new version of something, that version has come about for a reason. For `mocha` there is *a lot* that changed, mostly fixes (see [mocha/CHANGELOG.md@2.4.2](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#242--2016-01-26)), and I'd argue, too much to distill into a commit message (same for `eslint`).

lurch:
Why does your comment link to the `@2.4.2` section of the `mocha` changelog, when apparently we were already using mocha 2.4.5 ? \*confused\*

jviotti:
I added the label, let's see what happens :D

lekkas:
🎉

lurch:
@hedss @lekkas Are there "standard" colours for the procbots labels? I'm fairly sure that in most other repos they're purple...?
It's obviously only a minor point, but standard colours make things easier to spot at a glance.

lurch:
... and why did 4 out of 5 CI checks fail? Shouldn't that have stopped the merge?? Or has this repo been configured differently to other VersionBot-using repos? (and if so, is there a specific reason for that?)

jhermsmeier:
@lurch because the merge & branch deletion happens before the builds can start, causing `git fetch origin +refs/pull/176/merge` to fail, because the ref doesn't exist anymore.

hedss:
@lurch When it merged, nothing had failed. Another point, merges won't be prevented if the base hasn't been set up with a mandatory status check that fails it. Anything optional that doesn't actually stop the merge will be ignored.

lekkas:
@lurch currently we haven't set any specific rules for either colours or names of labels in general across resin.io . 'ready-to-merge' is a temporary label that is applied for a few minutes though, unlike others that are longer-lived (e.g. when categorising issues) and need to be more consistent . Till we have defined names/colours (that will ultimately have to be created using a cli tool) we can leave this up to the maintainer's taste :)

hedss:
I just picked the same colour all of the other resin added labels had, as it seemed logical to me at the time.

Issue #123, opened 2016-11-17
Display name of a drive


vohtaski:
Thanks for the nice module!

Would it be possible to display a device name?
For example, my usb drive has this name `PHY_RH_252`.
But I don't see it in the list of options
On mac:
```
description:"AT90USB128 M S"
device:"/dev/disk3"
mountpoints: {
    path:"/Volumes/PHY_RH_252"
    raw:"/dev/rdisk3"
size:3980394496
system:false
```

On windows:
```
description:"ATMEL AT90USB128 M S USB DEVICE"
device:"\\.\PHYSICALDRIVE1"
mountpoints:
    path:"D:"
    raw:"\\.\PHYSICALDRIVE1"
size:3980394496
system:false
```

jviotti:
Hi @vohtaski,

Drivelist lists the top level drives, and therefore includes the top level drive name in the `description` property, which is usually slightly different across platforms (depends on what the OS interprets the name), which in your example is "AT90USB128 M S" and "ATMEL AT90USB128 M S USB DEVICE".

"PHY_RH_252", in the other hand, is the name of one of the partitions. I'd love to have this functionality as `name` or `description` properties included in the mount-points array. 

How do you feel about trying a PR yourself? I can help with any OS you don't have access to.

vohtaski:
PR should be good. We definitely need it for the project at my work.
I just can't say for now if it's next week or couple weeks into the future :)

vohtaski:
@jviotti 
I finally managed to find some time to work on this issue.
My changes for windows and darwin are here.
Should be similar to linux but I dont have linux machine to test.
https://github.com/vohtaski/drivelist/commit/f9a8195df959821e901710a19e426529dd6b915f

I introduced a name `volumeName` next to `path` in `mountpoints` array.
Should I submit a PR?

jviotti:
Hi @vohtaski ,

Sorry for the delay! A PR would be very welcomed! When you create it, can you tick the option to allow the project maintainers to make changes to the PR? I can jump in and push Linux support there.

Issue #98, opened 2016-10-14
Improve "mountpoint" contents


feliupe:
Hi guys,

I am using Kubuntu and I am having a problem to get the my mountpoints, I am having just one value in the 'disks' array, having the 'mountpoint' value equal '/,/usr'. Maybe the right output should be something like: 

disks = [{mountpoint: '/', ...}, {mountpoint: '/usr', ...}]

Tell me if I can help with something. Thx

jviotti:
Hi @feliupe,

The purpose of `drivelist` is to list "top-level" disks (e.g: `/dev/sda`, `/dev/sdb`) rather than partitions, which is the reason a single top level disk may contain multiple partitions.

Ideally, we could make each drive contain a `partitions` property each including a different mount-point, but again, this is not the main purpose of `drivelist` and therefore not something we are prioritizing work for.

That being said, happy to take a PR of the improvement discussed above!

feliupe:
Thanks for your response @jviotti. I will consider about it. I've been taking a look on this part of the code and seems to not be that complex.

I will try to manage my time to work on it, but not sure. In this case should the issue be close?

jviotti:
Sounds great! Don't hesitate in pinging me if I can help with anything. Notice that the PR should implement this feature for all supported platforms. Happy to help/test on any platform you don't have access to.

Let's keep the issue opened until it gets implemented.

lurch:
I've just been wondering a very similar question (but from a slightly different viewpoint): instead of `mountpoint` being a single comma-separated string, would it be better for mountpoint to instead return a list of strings? Might even make sense to rename it to `mountpoints` in that case. E.g. `mountpoints: ['/', '/usr'],` ?  (which would also accommodate the highly-unlikely situation of a mountpoint itself containing a comma, which wouldn't work properly with the current code)

Obviously that breaks backwards-compatibility though, so would necessitate a version jump to 5.0.0

lurch:
In https://github.com/resin-io-modules/drivelist/issues/85#issuecomment-257383810 @jviotti said:
> Since changing the format of mountpoint is a breaking change, I suggest starting with the following:
```
[
  {
    "device": "/dev/sda",
    "mountpoints": [
      {
        "path": "/dev/sda1"
      },
      {
        "path": "/dev/sda2"
      }
    ],
    ...
  }
]
```

[...]

PR #177, opened 2017-06-19
add volumeName for each mountpoint

vohtaski:
Introduced a name `volumeName` next to path in `mountpoints` array. Addresses #123 for windows and mac. Linux is needed.


scripts/darwin.sh
```diff
@@ -61,6 +61,7 @@ for disk in $DISKS; do
     echo "mountpoints:"
     echo "$mountpoints" | while read -r mountpoint ; do
       echo "  - path: \"$mountpoint\""
+      echo "    volumeName: \"$mountpoint\""
```
lurch: Huh? You're setting both `path` and `volumeName` to the same `$mountpoint` variable??

scripts/win32.bat
```diff
@@ -128,12 +129,14 @@ Function GetTopLevelDrives()
 			Summary.Add "Size", TopLevelDrive.Size
 
 			Set Mountpoints = new List
+			Set MountpointObjects = new List
```
lurch: Could the code be tidied up to use _just_ `MountpointObjects`, and then get rid of `Mountpoints` ? (to avoid the duplication). I'm afraid I don't know VBScript very well myself.

scripts/win32.bat
```diff
@@ -145,7 +148,7 @@ Function GetTopLevelDrives()
 				End If
 			Next
 
-			Summary.Add "Mountpoints", Mountpoints
+			Summary.Add "Mountpoints", MountpointObjects
```
lurch: Perhaps the `"Mountpoints"` here (and on line 178) should be changed to `"MountpointObjects"` ?

lib/parse.js
```diff
@@ -68,7 +68,7 @@ module.exports = (input) => {
     device = _.chain(device)
       .split('\n')
       .filter((line) => {
-        return /^(\s\s-\s)?[a-z]+:/g.test(line);
+        return /^(\s\s[-\s]\s)?[a-zA-Z]+:/g.test(line);
```
lurch: Could you explain why this change was needed?

lurch:
And it would be great if this PR could update the examples in the README too :-)

lurch:
@hedss What's happening with `Versionist` in this PR's status-checks?

scripts/darwin.sh
vohtaski: On mac, `path` already has the name of the device `/Volumes/Device_115`.
Not sure what's the best here, probably to parse the string and get `Device_115`.

lib/parse.js
vohtaski: before parsing the output with yaml, this filter is applied, so it removes the line I added earlier.
"    volumeName: "

scripts/win32.bat
vohtaski: Probably yes, but I also don't know much VBScript.
There is this line in the code. I was not sure what it does
https://github.com/resin-io-modules/drivelist/blob/master/scripts/win32.bat#L135

`Mountpoints` is a list of strings and I introduced `MountpointObjects` which is a list of objects. But list function works only with simple types:
https://github.com/resin-io-modules/drivelist/blob/master/scripts/win32.bat#L37

So I used the `Mountpoints` for condition check, but store actual data inside `MountpointObjects`.

Not the best solution, imho, but was a minimal change in the code

scripts/darwin.sh
lurch: Is that _guaranteed_ though, or is it just common convention? Perhaps there's a more reliable way of getting the volumeName?

lurch:
Assigning this to @jviotti since (unfortunately for him) I think he knows more VBScript than any of the other Etcher team-members.

lib/parse.js
lurch: Why does it filter out the line? Don't you want the volumeName to also make it into the JSON objects?

scripts/darwin.sh
vohtaski: Looks like not `guaranteed`. I have the following drives
```
volumeName: "/"
volumeName: "/Volumes/My Passport"
```

lib/parse.js
vohtaski: yes, that's exactly what I want :)
But this line `/^(\s\s-\s)?[a-z]+:/g.test(line);` removes volumeName, since it's has uppercase and has 4 spaces before

vohtaski:
I've added a better way of extracting volume name on mac

scripts/darwin.sh
```diff
@@ -60,8 +60,11 @@ for disk in $DISKS; do
   else
     echo "mountpoints:"
     echo "$mountpoints" | while read -r mountpoint ; do
+      volume="$(echo "$mount_output" | perl -n -e'm{^('"${disk}"'(s[0-9]+)?) on (.*) \(.*\)$} && print "$1\n"')"
```
lurch: Could this be modified to not use `perl` ?
We wouldn't want to see https://github.com/resin-io/etcher/issues/1288 getting re-introduced ;-)

vohtaski: I think it can use sed or awk, but there is perl used just some lines above :)

jhermsmeier:
@lurch, I think versionist will only do anything if a label is applied, hence no status report from it yet

hedss:
@lurch @jhermsmeier   VersionBot only currently operates on Resin respositories. This is because it is not a public Integration. Unfortunately, in this case, we're trying to merge from an outside repo (`vohtaski:master`) and VersionBot has no access to it. So, we're going to need to rethink our workflow here. We could use a repo merge branch. I think this needs some discussion in a process call.

lib/parse.js
lurch: Ahhh, I think this line is used to filter out any line that _doesn't_ look like YAML, in order to try and prevent e.g. any error messages included in the script-output from breaking the YAML parser.
But according to both http://www.yaml.org/spec/1.2/spec.html#id2759963 and http://www.yamllint.com/ the example in `README.md` of:
```
device: /dev/disk1
description: "Macintosh HD"
size: 249779191808
mountpoints:
  - path: /
raw: /dev/rdisk1
protected: False
system: True
```
would be 'more correct' if formatted as:
```
device: /dev/disk1
description: "Macintosh HD"
size: 249779191808
mountpoints:
  -
    path: /
raw: /dev/rdisk1
protected: False
system: True
```
which I guess means (if the scripts were all updated to output this 'more correct' format) the regex could be updated to `/^((\s{4})?[a-z]+:)|(\s{2}-$)/`. And I guess you could change `volumeName` to something all-lowercase like `label` ;-)

Although having thought about it some more, I'm not sure if adding `label` to `mountpoints` is entirely what you want, as it'll only give you volume-labels for partitions that are actually mounted, and volume-labels for unmounted partitions won't be shown. So I think the discussion in this PR actually overlaps with that in #98

scripts/darwin.sh
lurch: Ah, good point! I'd forgotten that #168 only removes one usage of `perl`, not all usages of perl ;-)

oriolpuig:
Hi @vohtaski @lurch  and @jviotti,

Have you some update about it? 

I'm combining **drivelist** with **linux-mountutils** to mount and unmount in linux via NodeJS.

**Drivelist** reported me one device in `/dev/sdc` but If I do a  mount in bash, Linux says me `/dev/sdc1`. So, when I try to mount `/dev/sdc`, a **linux-mountutils** error displays "this device is already mounted or busy". 

That error is because drivelist doesn't provide the volumeName. 

Thanks for all your job.

BR

Oriol

vohtaski:
@oriolpuig 
I am using this fork for our needs, I don't know if there is a plan to merge it

lurch:
@oriolpuig `/dev/sdc` is the device name (i.e. the whole hard-drive) but `/dev/sdc1` is the partition name (i.e. the first (and maybe only?) partition on the drive). It's possible for a drive to contain multiple partitions. Etcher (which is what this module was written for) only needs the drive name (because it overwrites the whole drive at a low level). `linux-mountutils` (and the `mount` command) need the partition name, because they need to know which partition on the drive (there could be several) you actually want to mount.

oriolpuig:
yes, @lurch and that's why I would like to have the `volumeName`, to concat the `deviceName` and volumeName and mount it with the other library.

lurch:
@oriolpuig "volume name" is something else entirely, and I'm not sure it's actually what you're after. AFAIK volume name is effectively a text-label that you can add to a partition, _if_ the filesystem used on a partition supports such a feature.

lurch:
I guess this PR is now woefully out of date, now that the Windows part of `drivelist` is implemented as a native C++ module.

jhermsmeier:
Yeah, closing this, as it's not applicable anymore – yet we might want to look at adding that, if feasible.

PR #179, opened 2017-06-26
refactor(windows): implement volume::GetType()

jviotti:
This function uses GetDriveType() to fetch the volume type rather than
using WMI.


lurch:
Looking at https://msdn.microsoft.com/en-us/library/windows/desktop/aa364939(v=vs.85).aspx does this also work correctly for drives without volumes on them, e.g. an SD-card just filled with all zeroes?

jviotti:
> Looking at https://msdn.microsoft.com/en-us/library/windows/desktop/aa364939(v=vs.85).aspx does this also work correctly for drives without volumes on them, e.g. an SD-card just filled with all zeroes?

Yeah, it returns the correct type (e.g. removable or fixed) independently on the contents of the drive.

> EDIT: It seems frustrating how many Win32 API calls take a Volume or a Drive as a parameter, when what you actually want is information about the PhysicalDisk! I can only imagine how much more frustrating this must be for you Juanchi!

lol yeah, its very frustrating, but at least its all almost done, and we're just polishing the remaining rough edges.

PR #181, opened 2017-06-26
refactor(windows): Use an IOCTL to determine the partition type

jviotti:
This commit is part of some efforts to make use of IOCTLs directly and
get rid of WMI/COM altogether.

Change-Type: patch

### DIFF
--- a/src/windows/volume.cc
+++ b/src/windows/volume.cc
@@ -65,6 +65,41 @@ HRESULT drivelist::volume::GetReadOnlyFlag(const wchar_t letter, BOOL *out) {
   return result;
 }
 
+HRESULT drivelist::volume::HasFileSystem(const wchar_t letter, BOOL *out) {
+  HANDLE handle = drivelist::volume::OpenHandle(letter, 0);
+  if (handle == INVALID_HANDLE_VALUE)
+    return E_HANDLE;
+
+  PARTITION_INFORMATION_EX information;
+  DWORD bytesReturned;
+
+  if (!DeviceIoControl(handle, IOCTL_DISK_GET_PARTITION_INFO_EX, NULL, 0,
+                       &information, sizeof(information),
+                       &bytesReturned, NULL)) {
+    CloseHandle(handle);
+    return E_FAIL;
+  }
+
+  // See https://msdn.microsoft.com/en-us/library/windows/desktop/aa365448(v=vs.85).aspx
+  switch (information.PartitionStyle) {
+  case 0:  // MBR
+    *out = IsRecognizedPartition(information.Mbr.PartitionType);
+    break;
+  case 1:  // GPT
+
+    // TODO(jviotti): The Windows API documents a constant called
+    // PARTITION_ENTRY_UNUSED_GUID, but I can't find its definition anywhere.
+    *out = information.Gpt.PartitionType.Data1 != 0;
+
+    break;
+  default:  // RAW
+    *out = FALSE;
+  }
+
+  CloseHandle(handle);
+  return S_OK;
+}
+
 drivelist::volume::Type drivelist::volume::GetType(const wchar_t letter) {
   TCHAR drivePath[kVolumePathShortLength];
   sprintf_s(drivePath, "%c:\\", letter);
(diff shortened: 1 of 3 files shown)


jviotti:
They are used internally in order to decide what show or hide.

src/windows/volume.cc
```diff
@@ -65,6 +65,41 @@ HRESULT drivelist::volume::GetReadOnlyFlag(const wchar_t letter, BOOL *out) {
...
+
+  // See https://msdn.microsoft.com/en-us/library/windows/desktop/aa365448(v=vs.85).aspx
+  switch (information.PartitionStyle) {
+  case 0:  // MBR
+    *out = IsRecognizedPartition(information.Mbr.PartitionType);
+    break;
+  case 1:  // GPT
+
+    // TODO(jviotti): The Windows API documents a constant called
+    // PARTITION_ENTRY_UNUSED_GUID, but I can't find its definition anywhere.
```
lurch: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365449(v=vs.85).aspx says
```
PARTITION_ENTRY_UNUSED_GUID
00000000-0000-0000-0000-000000000000
```

src/windows/volume.cc
```diff
@@ -65,6 +65,41 @@ HRESULT drivelist::volume::GetReadOnlyFlag(const wchar_t letter, BOOL *out) {
   return result;
 }
 
+HRESULT drivelist::volume::HasFileSystem(const wchar_t letter, BOOL *out) {
```
lurch: Surely checking the partition type is something very different to a volume having a filesystem or not?
Consider creating a FAT32 partition on a disk (which updates the MBR) but then not formatting it. The volume has a valid partition type, but it doesn't have any filesystem yet...

IMHO either this function isn't doing what you think it is, or it should be renamed to something else...

src/windows/volume.cc
jviotti: Right, but as explained in the comment, that doesn't seem to exist:

```
..\src\windows\volume.cc(112): error C2065: 'PARTITION_ENTRY_UNUSED_GUID': undeclared identifier [C:\Projects\drivelist
\build\drivelist.vcxproj]
```

src/windows/volume.cc
jviotti: > Consider creating a FAT32 partition on a disk (which updates the MBR) but then not formatting it. The volume has a valid partition type, but it doesn't have any filesystem yet...

Oh yeah, I didn't think of that, and I can indeed trigger a bug in such scenario. How should we check if the partition is indeed there though?

src/windows/volume.cc
lurch: Oh, I see, I misinterpreted your comment ;-)
I just did a bit of searching, and found it it `diskguid.h`
```
DEFINE_GUID(PARTITION_ENTRY_UNUSED_GUID,    0x00000000L, 0x0000, 0x0000, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00);    // Entry unused
```

src/windows/volume.cc
jviotti: This is the error message I get:

> The volume does not contain a recognized file system. Please make sure that all required file system drivers are loaded and that the volume is not corrupt.

lurch: > How should we check if the partition is indeed there though?

I dunno, you know the Windows APIs much better than I do ;-)

lurch: The only thing vaguely relevant I could find was https://msdn.microsoft.com/en-us/library/windows/desktop/aa363785(v=vs.85).aspx

I suppose it's conceivable that the only way to check if a volume is properly formatted or not, might be to try mounting it (if it's not already mounted), and then catch any exception that gets thrown? \*shrug\*
Unless you want to use something like https://github.com/resin-io-modules/resin-image-fs but that might be overcomplicating things?

jviotti: After some research, looks like the only way to reliably detect this is to run `GetVolumeInformation` and check if the error is `ERROR_UNRECOGNIZED_VOLUME`

src/windows/volume.cc
jviotti: Awesome find. I ended up re-doing the whole function based on `GetVolumeInformation`, since that seems the only way to detect this.

PR #183, opened 2017-06-26
fix(windows): detect read-only flag on SD Cards without a filesystem

jviotti:
If an SD Card is locked, but contains no file-system,
GetVolumeInformation, opening a handle with write access, and even
IOCTL_DISK_GET_DISK_ATTRIBUTES would say that the volume is not locked.

The problem is that Windows makes a distinction between disks and
volumes, and a disk can be read-only even when its volumes are writable.

As a solution, we use a combination of IOCTL_DISK_IS_WRITABLE and
GetVolumeInformation().

See: https://github.com/resin-io/etcher/issues/1538
See: https://stackoverflow.com/a/44706760/1641422
Change-Type: patch

### DIFF
--- a/src/windows/scanner.cc
+++ b/src/windows/scanner.cc
@@ -214,15 +214,23 @@ ScanMountpoints(drivelist::com::Connection *const connection,
     if (FAILED(result))
       return InterpretHRESULT(result);
 
-    BOOL isProtected;
-    result = drivelist::volume::GetReadOnlyFlag(path[0], &isProtected);
+    // Turns out a disk can be writable while its volumes can be read-only
+    // and viceversa, so we need to check the writable capabilities of
+    // both the disk and its volumes
+    BOOL writable;
+    result = drivelist::volume::IsDiskWritable(path[0], &writable);
     if (FAILED(result))
       return InterpretHRESULT(result);
+    if (writable && hasFilesystem) {
+      result = drivelist::volume::IsVolumeWritable(path[0], &writable);
+      if (FAILED(result))
+        return InterpretHRESULT(result);
+    }
 
     drivelist::mountpoint_s mountpoint;
     mountpoint.path = path;
     mountpoint.disk = "\\\\.\\PHYSICALDRIVE" + std::to_string(number);
-    mountpoint.readonly = isProtected;
+    mountpoint.readonly = !writable;
     mountpoint.system = path[0] == systemDriveLetter;
     mountpoint.hasFilesystem = hasFilesystem;
     output->push_back(mountpoint);
--- a/src/windows/volume.cc
+++ b/src/windows/volume.cc
@@ -44,25 +44,45 @@ HRESULT drivelist::volume::GetDeviceNumber(const wchar_t letter, ULONG *out) {
   return S_OK;
 }
 
-HRESULT drivelist::volume::GetReadOnlyFlag(const wchar_t letter, BOOL *out) {
-  DWORD filesystem_flags = 0;
-  TCHAR drive_path[kVolumePathShortLength];
-  sprintf_s(drive_path, "%c:\\", letter);
+HRESULT drivelist::volume::IsDiskWritable(const wchar_t letter, BOOL *out) {
+  HANDLE handle = drivelist::volume::OpenHandle(letter, 0);
+  if (handle == INVALID_HANDLE_VALUE)
+    return E_HANDLE;
 
-  const HRESULT result =
-      GetVolumeInformation(drive_path, NULL, kVolumePathShortLength, NULL, NULL,
-                           &filesystem_flags, NULL, 0);
+  DWORD bytesReturned;
 
-  if (FAILED(result))
-    return result;
+  // The IOCTL_DISK_IS_WRITABLE returns
+  // FALSE if the device is read-only
+  // See https://msdn.microsoft.com/en-us/library/windows/desktop/aa365182(v=vs.85).aspx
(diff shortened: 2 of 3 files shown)


src/windows/scanner.cc
```diff
@@ -220,15 +220,23 @@ ScanMountpoints(drivelist::com::Connection *const connection,
...
+        return InterpretHRESULT(result);
+    }
 
     drivelist::mountpoint_s mountpoint;
     mountpoint.path = path;
     mountpoint.disk = "\\\\.\\PHYSICALDRIVE" + std::to_string(number);
-    mountpoint.readonly = isProtected;
+    mountpoint.readonly = !writable;
     mountpoint.system = path[0] == systemDriveLetter;
     mountpoint.hasFilesystem = hasFilesystem;
```
jhermsmeier: These are all not exposed on the JS side of things yet – is this intentional?

jviotti: We don't expose it to the end user, but we use it in `drivelist.cc` to determine which volumes to filter.

jhermsmeier: Shouldn't we leave this up to the module user? Filtering these would make using a readonly device as a source impossible, because it'd never show up, for example.

jviotti: I mean, we expose the read-only drive, but not the ones without a file-system, even though we still want to use some of its metadata for the top level disk.

For example, in Windows, we only determine if a top level drive is read-only if one of its volumes is read-only. There are drives that have no file-system, and even though Windows assigns it a letter in the background, is not something user can even see from the file explorer. By adding a boolean flag here, we can then check if its read-only or not, and set the top level property accordingly, but still filter it out from the final set of volumes.

jhermsmeier: > we only determine if a top level drive is read-only if one of its volumes is read-only.

Does this mean that devices with i.e. both NTFS & HFS+ partitions under Windows with Boot Camp will always be read-only, as the HFS+ driver is read-only on Windows with Boot Camp?

> I mean, we expose the read-only drive, but not the ones without a file-system,

So read-only devices with unrecognised file systems are filtered out as well? For example, a read-only SD card with an extfs on it?

jhermsmeier:
@jviotti this has conflicts that need resolving, btw

jviotti:
Conflicts resolved!

src/windows/scanner.cc
jviotti: > Does this mean that devices with i.e. both NTFS & HFS+ partitions under Windows with Boot Camp will always be read-only, as the HFS+ driver is read-only on Windows with Boot Camp?

In this case, we will show the `\\.\\PHYSICALDRIVE` entry, which will contain a single volume (e.g. `C:`) that corresponds to the NTFS partition

> So read-only devices with unrecognised file systems are filtered out as well? For example, a read-only SD card with an extfs on it?

Let's say a physical drive is detected as writable. This drive contains a read-only volume without a file-system. The output of drivelist will show a read-only `\\.\PHYSICALDRIVE` without any volumes.

So we filter out volumes without recognized file-systems, but we still take their writable state into consideration.

lurch: If a volume (volume == partition, right?) doesn't have a file-system, then how can it be read-only?

lurch: I'm having trouble understanding this thread, I suspect I'm getting confused between disks, drives, volumes, partitions, drive-letters, etc. :-/

jviotti: Windows still assigns volumes without file-systems a drive letter (even though UIs such as the file explorer filter them out), and they still can have a writable attribute.

jviotti: @lurch I'm super confused about it as well. Here's what I know so far:

- Drive -> `\\\\.\\PHYSICALDRIVEN`
- Disk -> `\\\\.\\X:`
- Volume -> `X:` (yes, they are different)
- Partitions = Volumes

lurch: How come the same `path[0]` parameter is being given to both `IsDiskWritable` and `IsVolumeWritable` ? Shouldn't we be checking that the _Drive_ is writable?

jviotti: Both `IOCTL_DISK_IS_WRITABLE` and `GetVolumeInformation()` take letters as arguments, but one of them returns the writable state of the disk, and the other returns the writable state of the volume. There doesn't seem to be any Windows function to determine the writable state of a drive.

Also, it looks like despite having the same identifier, a volume and a disk are different things, and each have their own copy of attributes such as this one.

I guess these kinds of awkwardness are what make the Windows APIs so confusing. Check https://stackoverflow.com/questions/44703463/getvolumeinformation-not-reporting-file-read-only-volume-for-a-locked-sd-card/44706760 for some more details.

jviotti: > And just to be clear, a mounted file system volume can be read-only even if the disk is writable so you need to check both if you want information about volumes and disks.

lurch: Looking again at the source code, I see that `IsVolumeWritable` does `sprintf_s(drivePath, "%c:\\", letter);` and `IsDiskWritable` calls `OpenHandle` which does `sprintf_s(devicePath, "\\\\.\\%c:", letter);`, so that at least answers part of my question ;-)

Does it mean that your Volume example above ought to be `X:\\` ?

jviotti: Oh, yeah, sorry, I'll edit the comment :)

Issue #126, opened 2016-11-29
Detect "operating system" drives


lurch:
As mentioned in https://github.com/resin-io/etcher/issues/888 it would be nice if we could identify drives where the currently-running operating system is loaded from. This would allow [Etcher](https://github.com/resin-io/etcher) to prevent writing to those drives, even if the user has enabled "Unsafe mode" (which allows you to write to "system" drives). (Unfortunately Removable Drives often get misidentified by `drivelist` as "system" drives)

We could do this by detecting drives containing partitions mounted at `/`, `/boot`, `/home`, etc. in the case of UNIX (and checking for `root=` in `/proc/cmdline` on Linux), and drives containing drive letters that match `%SYSTEMDRIVE%`, `%PROGRAMFILES%` and `%HOMEPATH%` in Windows.

See also https://github.com/resin-io/etcher/issues/1435

jviotti:
I think we can add a new property that is a subset of `system`. That is, if a drive is an "operating system drive", then its also `system: true`, but the opposite might not always be true.

BTW, this might be good opportunity to rename the `system` property to something more meaningful. What about renaming `system` to `removable`, and using `system` for the new "operating system drive" meaning?

lurch:
As long as changing the meaning of "system drive" won't be too confusing to Etcher's existing users...?

Other than Etcher, does `drivelist` have any other downstream projects that would be impacted by such a change?

jviotti:
> As long as changing the meaning of "system drive" won't be too
> confusing to Etcher's existing users...?

I don't think that will be a problem. We don't show "system drives" at
all yet (but we might show them all in the near future), and the only
place that refers to them is the unsafe mode modal, in which I guess a
minor wording change won't make a difference to users.

> Other than Etcher, does `drivelist` have any other downstream projects
> that would be impacted by such a change?

I know there are many modules depending on it (we can re-check NPM).
This would come as a major version increment, so compatibility shouldn't
be a problem. Since the new `removable` property will have the same
meaning as `system` does now, we can add an entry in the CHANGELOG
describing that to upgrade to the new version, simply change any
reference to `system` with `removable`.

On Tue, Nov 29, 2016 at 04:29:38PM -0800, Andrew Scheller wrote:
> As long as changing the meaning of "system drive" won't be too confusing to Etcher's existing users...?
> 
> Other than Etcher, does `drivelist` have any other downstream projects that would be impacted by such a change?
> 
> -- 
[...]

PR #184, opened 2017-06-28
refactor(windows): don't use WMI to list volumes

jviotti:
Change-Type: patch

### DIFF
--- a/src/windows/scanner.cc
+++ b/src/windows/scanner.cc
@@ -143,73 +143,33 @@ static drivelist::Code ScanDisks(drivelist::com::Connection *const connection,
   return drivelist::Code::SUCCESS;
 }
 
-static HRESULT
-GetOperatingSystemDriveLetter(drivelist::com::Connection *const connection,
-                              wchar_t *out) {
-  drivelist::wmi::Query query(L"SELECT SystemDrive FROM Win32_OperatingSystem");
-  HRESULT result = query.Execute(connection);
-  if (FAILED(result))
-    return result;
-
-  while (query.HasResult()) {
-    result = query.GetPropertyCharacter(L"SystemDrive", out);
-    if (FAILED(result))
-      break;
-    result = query.SelectNext();
-    if (FAILED(result))
-      break;
-  }
-
-  return result;
-}
-
 static drivelist::Code
-ScanMountpoints(drivelist::com::Connection *const connection,
-                std::vector<drivelist::mountpoint_s> *const output) {
+ScanMountpoints(std::vector<drivelist::mountpoint_s> *const output) {
   wchar_t systemDriveLetter;
-  HRESULT result =
-      GetOperatingSystemDriveLetter(connection, &systemDriveLetter);
+  HRESULT result = drivelist::volume::GetSystemVolume(&systemDriveLetter);
   if (FAILED(result))
     return InterpretHRESULT(result);
 
-  drivelist::wmi::Query query(L"SELECT * FROM Win32_Volume");
-  result = query.Execute(connection);
+  std::vector<wchar_t> volumes;
+  result = drivelist::volume::GetAvailableVolumes(&volumes);
   if (FAILED(result))
     return InterpretHRESULT(result);
 
-  BSTR string;
-  ULONG number;
-
-  while (query.HasResult()) {
-    result = query.GetPropertyString(L"DriveLetter", &string);
-    if (FAILED(result))
-      return InterpretHRESULT(result);
-    if (string == NULL) {
-      result = query.SelectNext();
-      if (FAILED(result))
-        return InterpretHRESULT(result);
-      continue;
-    }
-
-    std::string path = ConvertBSTRToString(string);
-    SysFreeString(string);
-
-    BOOL hasFilesystem;
-    result = drivelist::volume::HasFileSystem(path[0], &hasFilesystem);
-    if (FAILED(result))
-      return InterpretHRESULT(result);
-
+  for (wchar_t volume : volumes) {
     // We only want to consider removable or local disks in this module
-    drivelist::volume::Type volumeType = drivelist::volume::GetType(path[0]);
+    drivelist::volume::Type volumeType = drivelist::volume::GetType(volume);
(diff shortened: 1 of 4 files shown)


src/windows/scanner.cc
```diff
@@ -144,100 +144,56 @@ static drivelist::Code ScanDisks(drivelist::com::Connection *const connection,
...
         return InterpretHRESULT(result);
     }
 
     drivelist::mountpoint_s mountpoint;
-    mountpoint.path = path;
+    mountpoint.path = std::string(1, static_cast<char>(volume)) + ":";
     mountpoint.disk = "\\\\.\\PHYSICALDRIVE" + std::to_string(number);
     mountpoint.readonly = !writable;
-    mountpoint.system = path[0] == systemDriveLetter;
+    mountpoint.system = volume == systemDriveLetter;
```
lurch: I realise I'm probably commenting on the 'wrong' PR for this, but please bear with me as I'm trying to get caught up with everything, but out-of-order... ;-)

Is this line correct? It means there'll only ever be one "system" drive on windows, whereas up until now drivelist has used "system" to indicate fixed, non-removable drives??
I agree that we _should_ detect the drive that the OS has booted from, but I think it'll be confusing to change the meaning of the `system` property, which we've already been using for other purposes.

src/windows/volume.cc
```diff
@@ -18,6 +18,32 @@
...
+  DWORD logicalDrivesMask = GetLogicalDrives();
+  if (logicalDrivesMask == 0)
+    return E_FAIL;
+
+  TCHAR currentDriveLetter = 'A';
+
+  while (logicalDrivesMask) {
+    if (logicalDrivesMask & 1)
+      output->push_back(currentDriveLetter);
+    currentDriveLetter++;
```
lurch: Should we check that this doesn't end up giving us drive-letters beyond 'Z` ?

src/windows/volume.cc
```diff
@@ -18,6 +18,32 @@
 
 static const size_t kVolumePathShortLength = 4;
 
+HRESULT drivelist::volume::GetSystemVolume(wchar_t *out) {
+  const char * const systemDrive = getenv("SystemDrive");
```
lurch: I wonder if there's a more reliable way of determining this, than checking a particular environment variable?
After following a bunch of deprecated-links from https://stackoverflow.com/questions/810273/how-can-i-get-the-system-drive-letter I eventually ended up at https://msdn.microsoft.com/en-us/library/bb762188.aspx and https://msdn.microsoft.com/en-us/library/dd378457(v=vs.85).aspx which I guess suggests we could check `FOLDERID_System` ?

lurch:
But just to clarify (I didn't realise this at first), this PR _does_ still use WMI for listing disks, it only removes the usage of WMI for listing the mountpoints (volumes) on those disks.

@jviotti I wonder if it'd make sense to remove the usage of WMI in ScanDisks and ScanMountpoints in the same PR?

lurch:
@hedss Is versionbot going funny again, or is it just more GitHub hiccups?

jviotti:
> @jviotti I wonder if it'd make sense to remove the usage of WMI in ScanDisks and ScanMountpoints in the same PR?

I'd love to, but that would take an insane amount of work. I guess its better to go incrementally.

src/windows/scanner.cc
jviotti: No worries! So this is part of the Windows disk/volume/drive mess. In this particular part of the code, we're checking of a mountpoint (e.g. volume) is a system one or not. We can only check if a **drive** is removable or not, so at this point, we can only check the letter against the operating system volume.

When joining both the volumes and the drives (in `drivelist.cc`), we take both into consideration, in order to pass to JavaScript the final "system" value we used to have before.

src/windows/volume.cc
jviotti: We could, but I don't think its necessary. If we go past `Z`, `logicalDrivesMask` will be `false`, and thus the `while` loop will finish.

src/windows/scanner.cc
lurch: Ah. To prevent confusion (like I just had!) maybe it'd better to use an attribute-name other than `system` here?

lurch: ...and given that you're determining here whether a disk/volume/drive is `REMOVABLE_DISK` or `LOCAL_DISK`, would it be worth storing that info on the `mountpoint` object too?

src/windows/volume.cc
jviotti: I implemented the suggestion from https://github.com/resin-io-modules/drivelist/issues/126#issuecomment-312071764

src/windows/scanner.cc
jviotti: > Ah. To prevent confusion (like I just had!) maybe it'd better to use an attribute-name other than system here?

I guess `system` makes a lot of sense in this case, so maybe worth changing the other `system` property to `removable`?

> ...and given that you're determining here whether a disk/volume/drive is REMOVABLE_DISK or LOCAL_DISK, would it be worth storing that info on the mountpoint object too?

I'm determining this for disks, not for mountpoints, so I wouldn't be able to add that to the mountpoint object.

jviotti: Ah lol, the `disk_s` `struct` already had a `removable` property instead of a `system` one

lurch: See, I told you I found all of this confusing... ;-)

hedss:
There's a conflict here for VB. It's an `unmerged` state because version `v5.0.25` already exists on `master`. Rebasing should work, as it'll pull in the new `CHANGELOG.md`, and VB will start afresh with a new version.

lurch:
I think this was caused by several PRs having the `ready-to-merge` label applied while each of them was waiting for the CI tests to finish, so it was VersionBot itself that bumped each PR to the same version `v5.0.25`. Couldn't VersionBot resolve this situation itself automatically?
At the very least it'd be nice if it could a) print a more helpful error message and b) not spam the PR with the same message over and over again.

hedss:
@lurch The coming changes to how bumps occur will fix this. As long as everything's fine when the merge is started, merges will happen instantly, instead of waiting for checks to pass again.

PR #185, opened 2017-06-28
fix(windows): all drives are interpreted as read-only

jviotti:
This was an accidental subtle logical flaw in the PR where we check both
the disk and the volume writable state.

We only want to consider the volumes writable state if the disk is
read-only, and the volumes have a file-system.

Change-Type: patch
See: https://github.com/resin-io-modules/drivelist/pull/183

### DIFF
--- a/src/windows/volume.cc
+++ b/src/windows/volume.cc
@@ -86,38 +86,26 @@ HRESULT drivelist::volume::IsVolumeWritable(const wchar_t letter, BOOL *out) {
 }
 
 HRESULT drivelist::volume::HasFileSystem(const wchar_t letter, BOOL *out) {
-  HANDLE handle = drivelist::volume::OpenHandle(letter, 0);
-  if (handle == INVALID_HANDLE_VALUE)
-    return E_HANDLE;
-
-  PARTITION_INFORMATION_EX information;
-  DWORD bytesReturned;
-
-  if (!DeviceIoControl(handle, IOCTL_DISK_GET_PARTITION_INFO_EX, NULL, 0,
-                       &information, sizeof(information),
-                       &bytesReturned, NULL)) {
-    CloseHandle(handle);
-    return E_FAIL;
+  TCHAR drivePath[kVolumePathShortLength];
+  sprintf_s(drivePath, "%c:\\", letter);
+  BOOL result = GetVolumeInformation(drivePath, NULL, 0, NULL,
+                                     NULL, NULL, NULL, 0);
+  if (result) {
+    *out = TRUE;
+    return S_OK;
   }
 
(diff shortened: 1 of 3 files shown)


lurch:
Huh? If the disk itself is read-only, then how can it have any writable volumes?

jviotti:
@lurch Windows weirdness :P

lurch:
Hmm, so `IsDiskWritable` might return False, even if the disk **is** actually writable? :confused:

jviotti:
Yeah, turns out we have to check the writable state of both the disk and its volumes to come up with an accurate result.

This is what I got after quite a lot of experimentation, and it looks like the logic yields correct results, at least on all the drives I tested.

lurch:
Doesn't seem to make any sense, but if you've tested it and it all seems to work okay, then I'll have to assume that it's correct ;-)  Stupid Windows APIs :crying_cat_face: 

Given the changes that #183 made to `src/windows/volume.h` should this be change-type: minor instead of change-type: patch ?

lurch:
Have you also checked SD cards with no volumes (partitions) on them at all, i.e. just filled entirely with zeroes? (and of course with the physical SD card switch in both the locked and unlocked positions)

jviotti:
I tested your suggestion, which revealed the issue in `HasFilesystem` that we were talking about in the other PR. I added a commit here that fixes everything!

src/windows/volume.cc
```diff
@@ -86,38 +86,26 @@ HRESULT drivelist::volume::IsVolumeWritable(const wchar_t letter, BOOL *out) {
...
+  sprintf_s(drivePath, "%c:\\", letter);
+  BOOL result = GetVolumeInformation(drivePath, NULL, 0, NULL,
+                                     NULL, NULL, NULL, 0);
+  if (result) {
+    *out = TRUE;
+    return S_OK;
   }
 
-  // See https://msdn.microsoft.com/en-us/library/windows/desktop/aa365448(v=vs.85).aspx
-  switch (information.PartitionStyle) {
```
lurch: It might still be worth checking the partition-info, before checking the filesystem with GetVolumeInformation? Since I guess if the partition info is invalid, we know straight away that the filesystem will be invalid too?

jviotti: I rather use a single function call than 2 in the worst case. Using just `GetVolumeInformation` also makes it all much more readable.

lurch: Fair enough.

lurch:
Poor VersionBot, getting all confused ;-)

jviotti:
@hedss There seems to be something wrong here. VersionBot complains, but this branch is completely up to date.

jviotti:
I'll merge this manually so we can keep moving.

Issue #142, opened 2017-02-18
Method of getting drive size on Windows needs changing


lurch:
For all the gory details, see my comments in https://github.com/resin-io/etcher/issues/995

But in summary: `scripts/win32.bat` in this repo currently gets the size of the drive using [Win32_DiskDrive](https://msdn.microsoft.com/en-us/library/aa394132(v=vs.85).aspx) which erroneously calculates the size of the raw drive using the CHS values. While this works fine for mechanical hard drives, it doesn't work for flash media (SD cards, USB sticks, etc.) and often under-reports the actual size of the drive. [IOCTL_DISK_GET_DRIVE_GEOMETRY_EX](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365171(v=vs.85).aspx) needs to be used instead, to get the _actual_ size of the raw drive.

I believe that fixing this will fix all the problems mentioned in https://github.com/resin-io/etcher/issues/995

jviotti:
Fixed by https://github.com/resin-io-modules/drivelist/pull/186

PR #186, opened 2017-06-28
fix(windows): report a more accurate drive size number

jviotti:
We were using WMI to get the size of a physical drive, however turns out
that is number is not very accurate (since its based on CHS), and even
plain wrong in some cases.

As a solution, we now use the IOCTL_DISK_GET_DRIVE_GEOMETRY_EX ioctl to
get the size of a physical drive.

See: https://github.com/resin-io/etcher/issues/995
See: https://github.com/resin-io/etcher/issues/1483
Change-Type: patch

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -14,6 +14,7 @@
             "src/windows/com.cc",
             "src/windows/scanner.cc",
             "src/windows/volume.cc",
+            "src/windows/disk.cc",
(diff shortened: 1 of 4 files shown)


jviotti:
Damn, I forgot about versionbots.

lurch:
So presumably this fixes #142 ? ;-P

PR #187, opened 2017-06-29
chore(package): Fix missing EOL in package.json

jhermsmeier:
This adds the missing EOL at the end of the `package.json`

Change-Type: patch


jviotti:
Merging myself, since VersionBot went nuts.

Issue #190, opened 2017-07-05
Avoid thrashing tmp directory with scripts


jhermsmeier:
See https://github.com/resin-io/etcher/issues/1571 for a detailed report

PR #191, opened 2017-07-05
fix(execute): Properly clean up tmp file after exec

jhermsmeier:
This fixes temporary files not being cleaned up after execution.
See https://github.com/resin-io/etcher/issues/1571 for extensive details.

Changes:

- Use `os.tmpdir()` to get temporary file directory
- Prefix tmp file with package name for better identification
- Delete tmp file after use
- Add execution timeout of 10s
- Add debug statements for write & unlink
- Add process signal to error message
- Remove `tmp` and `bluebird` dependencies

Change-Type: patch
Connects To: #190, https://github.com/resin-io/etcher/issues/1571

### DIFF
--- a/lib/execute.js
+++ b/lib/execute.js
@@ -16,13 +16,30 @@
 
 'use strict';
 
-const debug = require('debug')(require('../package.json').name);
-const Bluebird = require('bluebird');
-const fs = Bluebird.promisifyAll(require('fs'));
+const packageInfo = require('../package.json');
+const debug = require('debug')(packageInfo.name);
+const fs = require('fs');
 const childProcess = require('child_process');
 const path = require('path');
-const tmp = require('tmp');
-tmp.setGracefulCleanup();
+const os = require('os');
+const crypto = require('crypto');
+
+/**
+ * @summary Generate a tmp filename with full path of OS' tmp dir
+ * @function
+ * @private
+ *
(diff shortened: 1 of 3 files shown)


lib/execute.js
```diff
@@ -86,35 +104,40 @@ const createTemporaryScriptFile = (extension, contents) => {
...
+
   const extension = path.extname(script.originalFilename);
-  createTemporaryScriptFile(extension, script.content).then((temporaryPath) => {
-    return new Bluebird((resolve, reject) => {
 
-      // On Windows, if the directory containing the temporary
-      // path includes an ampersand (e.g. if the username contains
-      // an ampersand) then `.execFile` will fail, even with proper
-      // quoting, which is why we're using `.exec` instead.
-      return childProcess.exec(`"${temporaryPath}"`, (error, stdout, stderr) => {
```
jhermsmeier: This isn't required anymore, as Windows already uses the native bindings.

lurch:
Are there potentially any other 'temporary scripts' that we spawn anywhere, that might suffer from this same problem?

jhermsmeier:
No, not in this module, or do you mean elsewhere in Etcher?

lurch:
Yeah, I meant elsewhere in Etcher ;-)  (and the various modules it uses, that we've written)

PR #192, opened 2017-07-06
feat(windows): print error code string description in case of errors

jviotti:
Change-Type: minor

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -6,7 +6,9 @@
         "<!(node -e \"require('nan')\")",
         "."
       ],
-      "sources": [],
+      "sources": [
+        "src/code.cc"
+      ],
       "conditions": [
         [ 'OS=="win"', {
           "sources": [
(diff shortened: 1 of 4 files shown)


jviotti:
Fixed!

jviotti:
@hedss @lekkas I don't know what's going on, but VersionBot is not working on this repo anymore :(

hedss:
@jviotti The release tag for `5.0.25` isn't valid, it should be `v5.0.25` so `versionist` can't find the reference. I guess it was added manually.

jviotti:
@hedss I see, I always seem to get the `v` prefix wrong :P Could we make VersionBot print a more descriptive error message, or maybe even a warning, but still make it cope with the lack of the `v`?

lurch:
I think if we used VersionBot 100%, and never merged manually, then we'd never get invalid tags?

jviotti:
True. I went for manual merges given that we needed to keep moving, and I didn't know how to fix it.

hedss:
It's a fair comment. Ideally `versionist` needs to be libraried so we can can use it and get better error messages out. Unfortunately I've had zero time to do this. :-(

PR #193, opened 2017-07-06
feat(windows): print debugging information when DRIVELIST_DEBUG is set

jviotti:
Like we do on mountutils. Debugging native modules can be very tricky.
Printing a lot of info should help us to easily identify and fix issues.

Change-Type: minor

### DIFF
--- a/src/windows/com.cc
+++ b/src/windows/com.cc
@@ -15,16 +15,17 @@
  */
 
 #include "src/windows/com.h"
+#include "src/log.h"
 
 HRESULT drivelist::com::Initialize() {
   HRESULT result;
 
-  // Initialize COM
+  drivelist::Debug("Initializing COM");
   result = CoInitializeEx(NULL, COINIT_MULTITHREADED);
   if (FAILED(result))
     return result;
 
-  // Set general COM security levels
+  drivelist::Debug("Initializing COM security levels");
   result = CoInitializeSecurity(
     NULL,                         // Security descriptor
     -1,                           // COM authentication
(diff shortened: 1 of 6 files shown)


src/windows/com.cc
```diff
@@ -15,16 +15,17 @@
  */
 
 #include "src/windows/com.h"
+#include "src/log.h"
 
 HRESULT drivelist::com::Initialize() {
   HRESULT result;
 
-  // Initialize COM
+  drivelist::Debug("Initializing COM");
```
lurch: The COM stuff is just because we're still using WMI isn't it, and so it'll all go when we've finally got rid of all the code using WMI (by replacing it with lower-level APIs)?

src/windows/scanner.cc
```diff
@@ -146,29 +156,35 @@ static drivelist::Code ScanDisks(drivelist::com::Connection *const connection,
 static drivelist::Code
 ScanMountpoints(std::vector<drivelist::mountpoint_s> *const output) {
   wchar_t systemDriveLetter;
+  drivelist::Debug("Getting system volume drive letter");
   HRESULT result = drivelist::volume::GetSystemVolume(&systemDriveLetter);
   if (FAILED(result))
     return InterpretHRESULT(result);
```
lurch: Perhaps it'd have been useful here to actually log _which_ `systemDriveLetter` we got? ;)

lurch:
Any idea what went wrong with VersionBot / versionist here?

src/windows/com.cc
jviotti: Yeah, it'll get removed very soon.

jviotti:
Looks like I created a `5.0.25` tag instead of `v5.0.25`, which confused VersionBot. It should be all fixed now!

PR #194, opened 2017-07-17
perf: re-use existing temporary scripts when possible

jviotti:
This commit make drivelist attempt to re-use existing extracted scripts
instead of writing a new file each time for performance reasons.

For security reasons, we check that the script contents match the MD5
hash of the original scripts before running it.

Change-Type: minor


lib/execute.js
```diff
@@ -16,68 +16,149 @@
...
+    }
+
+    if (!isFile) {
+      debug('Creating script for the first time', output);
+      return createExecutableFile(script.content, output, callback);
+    }
+
+    // Integrity check, so we don't expose a vulnerability by letting
+    // other processes make this module run arbitrary code.
+    checksum.calculateFromFile(script.checksumType, output, (checksumError, scriptChecksum) => {
```
lurch: You could also check that the file has the correct size before calculating the checksum, as that'll allow you discard a possible mis-match much faster?

lib/execute.js
```diff
@@ -16,68 +16,149 @@
...
+      if (checksumError) {
+        return callback(checksumError);
+      }
+
+      if (script.checksum === scriptChecksum) {
```
lurch: Is there still a possible timing-attack here, where the file's contents are modified between verifying the checksum and executing the script? Pinging @petrosagg

lurch:
I guess there's also a trade-off as to whether continually reading and checksumming a file, is _actually_ any faster than deleting and recreating it (with a random filename) each time?

lib/execute.js
jviotti: Here's an alternative: we could sign the scripts and store the signed buffers in the JSON file, and check the signatures instead.

jviotti:
@lurch Right. Looks like the gains are very very small. Removing the checksum checks gives us ~60% faster execution times. I'll keep researching...

jviotti:
Code signing seems to also be a small speed gain on macOS, but we're basically discussing milliseconds from now on, which makes me think this particular task its not worth keep optimizing for.

jviotti:
Still, the timing attack vulnerability we're talking about is still present in the current implementation, so I guess its still worth discussing that, maybe in an arch call?

@petrosagg For context, the issue is that this module creates temporary scripts and executes them in order to scan the drives. In Etcher, those scripts are executed with root privileges in the writer process right before the flashing starts, which means that if an attacked with user access to the computer can predict the location where those scripts will be extracted, he can modify them with user privileges, and they will be run as root. The timing window is very small (micro-seconds), but still sounds like something we should consider.

jviotti:
We could refactor the writer process to not need to spawn these scripts, which fixes the `root` vulnerability, but there is still the issue that any user can make any other user run arbitrary commands.

lurch:
...and just as a reminder, here's the reason why we do things in this way: https://github.com/resin-io-modules/drivelist/pull/164

lib/execute.js
jhermsmeier: @lurch that timing attack exists regardless of the mechanics around it, even if you write out the file & immediately try to execute it.

lurch: True. But if somebody _did_ want to use this as an 'attack vector', I guess it's slightly easier if the file has a consistent name from one execution to the next, rather than getting a random name each time? ;)

jhermsmeier: They all have the same prefix, so that's irrelevant too. Either way, if something is on your machine, it's game over already – not sure if this vector makes sense worrying about in the first place.

lurch: The bit that _is_ worth worrying about here (which I guess is also a problem with the current version of Etcher), is that the scripts get written out with user-level access, but then executed by a process with root-level access.

jviotti:
Let's close this, and instead focus on getting rid of temporary scripts.

PR #195, opened 2017-07-18
fix: handle EAGAIN spawning errors

jviotti:
EAGAIN errors come from `uv_spawn()`, and may happen because of the
following reasons (from the man pages):

- The system-imposed limit on the total number of processes under
  execution would be exceeded. This limit is configuration-dependent

- The system-imposed limit MAXUPRC (<sys/param.h>) on the total number
  of processes under execution by a single user would be exceeded

This commit retries the spawning operation various times before giving
up with a human friendly error message.

See: https://github.com/resin-io/etcher/issues/1578

### DIFF
--- a/lib/execute.js
+++ b/lib/execute.js
@@ -69,20 +69,54 @@ const createTemporaryScriptFile = (extension, contents, callback) => {
 };
 
 /**
- * @summary Delete a file while ignoring any errors
+ * @summary Execute a script
  * @function
  * @private
  *
- * @param {String} filename - file path
- * @param {Function} callback - callback
+ * @param {String} scriptPath - script path
+ * @param {Function} callback - callback (error, output)
+ * @param {Number} [times=5] - retry times
+ *
+ * @example
+ * executeScript('path/to/script.sh', (error, output) => {
+ *   if (error) throw error;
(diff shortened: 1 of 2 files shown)


jviotti:
I'm merging this so we can release the next Etcher version soon. I'm happy to address any comments later.

jviotti:
Ah, damn, I should have used VersionBots :(

lib/execute.js
```diff
@@ -69,20 +69,54 @@ const createTemporaryScriptFile = (extension, contents, callback) => {
...
-const deleteTempFile = (filename, callback) => {
-  fs.unlink(filename, (error) => {
+const executeScript = (scriptPath, callback, times = 5) => {
+  childProcess.execFile(scriptPath, (error, stdout, stderr) => {
     if (error) {
-      debug('unlink', error);
+      if (error.code === 'EAGAIN') {
+        if (times <= 0) {
+          error.message = [
+            'Looks like you hit the limit of the number of process you can have at any given time,',
```
lurch: nit: process -> processes

jviotti: See https://github.com/resin-io-modules/drivelist/pull/197

PR #196, opened 2017-07-18
refactor: reduce debug logging if unlink was successful

jviotti:
As a way to reduce the amount of debug logging this module produces, we
only emit an unlink debug message if the unlink operation was succesful.

Also, as a better way to correlate unlink issues with a certain
drivelist call, we only report back the result of this module after the
unlink operation ended (sucessfully or not).

See: https://github.com/resin-io/etcher/pull/1600

### DIFF
--- a/lib/execute.js
+++ b/lib/execute.js
@@ -74,10 +74,15 @@ const createTemporaryScriptFile = (extension, contents, callback) => {
  * @private
  *
  * @param {String} filename - file path
+ * @param {Function} callback - callback
  */
-const deleteTempFile = (filename) => {
+const deleteTempFile = (filename, callback) => {
   fs.unlink(filename, (error) => {
-    debug('unlink', error || 'OK');
+    if (error) {
+      debug('unlink', error);
+    }
+
+    callback();
   });
 };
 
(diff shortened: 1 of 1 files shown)


lurch:
What does this VB error mean @hedss ?

lurch:
@jviotti "we only emit an unlink debug message if the unlink operation was succesful." Presumably you meant "...was unsuccessful" ? ;-)

hedss:
Strangely it appears that VB couldn't access the repo. I've just tried my test repo, and all is well. I suspect this might have been a GH wobble, as there's nothing else in the logs to suggest VB wasn't trying to communicate with it properly.

lurch:
So I guess if this happens again, the fix would be to re-kick VersionBot by adding a label or something?
I wonder if there could be a 'standard' label we could use to signify to VB that we want it to take another look at a PR, with VB then automatically removing that label after looking at the PR? Is that worth creating a suggestion issue for, or will this scenario be covered by https://github.com/resin-io-modules/resin-procbots/issues/196 anyway?

hedss:
Yeah, that's what https://github.com/resin-io-modules/resin-procbots/issues/196 is for, as you say. Though I'm not overly happy about it...

lurch:
Are the logs kept by VB detailed enough for you to be able to ask GH support why their API keeps having "hiccups" ?

PR #199, opened 2017-07-30
feat: add a `displayName` property to drive objects

jviotti:
The display name property is equal to the device path in macOS and
GNU/Linux, but its equal to the list of volumes with file-systems on
Windows, for user-friendliness purposes.

Change-Type: minor


jviotti:
This is something we were manually adding to the result of `drivelist` on Etcher, and its tied to https://github.com/resin-io/etcher/pull/1630

lurch:
Might be nice if the README could show an example of a disk on Windows with multiple drive-letters per disk? (otherwise the fact that this `.name` property might contain a comma-separated list is entirely unclear)

And I wonder if calling it something like `.display_name` would make it clearer that this property is just for display purposes, and you shouldn't really use it for anything else? (given that it has different meanings on different platforms)

jviotti:
> Might be nice if the README could show an example of a disk on Windows with multiple drive-letters per disk? (otherwise the fact that this .name property might contain a comma-separated list is entirely unclear)

Sounds good. I'll do it in a second.

> And I wonder if calling it something like .display_name would make it clearer that this property is just for display purposes, and you shouldn't really use it for anything else? (given that it has different meanings on different platforms)

Fair enough, lets go for it!

src/windows/scanner.cc
```diff
@@ -245,6 +246,22 @@ drivelist::Scanner::Scan(std::vector<drivelist::disk_s> *const output) {
 
   for (drivelist::disk_s disk : disks) {
     disk.mountpoints = mountpointMap[disk.id];
+
+    // For user friendly purposes, in Windows, the disk name is
+    // simply the list of drive letters, separated by commas.
+    disk.displayName = std::accumulate(
```
jviotti: Uh, I just realized I'm not handling the case where there are no mountpoints :/ I'll fix it in a minute

lurch: Good catch! Could you also add an example of that to the README please?

lib/scripts.json
```diff
@@ -1,12 +1,12 @@
...
     "originalFilename": "linux.sh",
     "type": "text"
   }
-}
+}
```
jhermsmeier: Missing EOL here

lib/parse.js
```diff
@@ -68,7 +68,7 @@ module.exports = (input) => {
     device = _.chain(device)
       .split('\n')
       .filter((line) => {
-        return /^(\s\s-\s)?[a-z]+:/g.test(line);
+        return /^(\s\s-\s)?[a-zA-Z]+:/g.test(line);
```
jhermsmeier: This can be changed to `/^(\s{2}-\s)?[a-z]/i` - no need for a global match flag when testing for the start of a string

src/windows/scanner.cc
jviotti: Done!

lib/scripts.json
jviotti: Fixed!

lib/parse.js
jviotti: Good catch! 

> Also, why is this testing for \s\s-\s, when none of the example input has that prefix?
Is that something that one of the scripts produces?

Mountpoints have that prefix. For example:

```yaml
device: /dev/disk1
displayName: /dev/disk1
description: "Macintosh HD - APPLE SSD SM0256G"
size: 140999917568
mountpoints:
  - path: "/"
raw: /dev/rdisk1
protected: False
system: True
```

jhermsmeier: Ah, missed that

lib/scripts.json
lurch: Given that this file is auto-generated anyway, how did this 'mistake' creep in?

jviotti: Generating the file from Windows messed things up :(

Issue #198, opened 2017-07-20
Error: Couldn't initialize the scanner: An unknown error occurred


barchard:
Hi, I have started receiving this error on Win10 (Creators Update) in Node + Electron. I am getting this error when calling `.list()` method in order to list devices. Strangely, this was previously working. Obviously, the error is vague so I'm not sure how to proceed. Any advice is appreciated. Thanks

barchard:
I should add, reverting to `v5.0.22` fixes the issue. I will find out which version this first appears in.

barchard:
NPM wouldn't allow me to install `v5.0.23` or `v5.0.24`. It said the only versions available are `v5.0.27`, `v5.0.26`, `v5.0.25`, and `v5.0.22`. The error appears to first appear in `v5.0.25`

rswamy:
I am setting up a new instance of my nodejs app (today) and was receiving some strange errors with the drivelist module when trying to run it. Googling led me to this issue somehow. My app fails with v5.0.27 so I reverted to v5.0.25 which I knew worked from a previous instance. Can't contribute more than this anecdote, I will need to look at what changed between these versions...

jviotti:
Hi @barchard @rswamy ,

> NPM wouldn't allow me to install v5.0.23 or v5.0.24. It said the only versions available are v5.0.27, v5.0.26, v5.0.25, and v5.0.22. The error first appears in v5.0.25.

Sorry about that. We're adopting an company wide technology to handle automated versioning and deployment, and things didn't go very smoothly on certain versions, which have not been published to npm.

***

The new drivelist version contains a very large change, where we re-wrote the Windows drive scanning routines in C++, to workaround a bug coming from WMI itself, where the reported drive sizes were not accurate for some drives.

Can you guys try to re-run drivelist with the `DRIVELIST_DEBUG` variable set? For example:

```sh
git clone https://github.com/resin-io-modules/drivelist
cd drivelist
npm install
set DRIVELIST_DEBUG=1
node example
```

That should provide debugging information so we can pin-point the issue.

barchard:
When I do this, all I see in the command prompt is:

```
[drivelist] Initializing scanner
[drivelist] Initializing COM
[drivelist] Initializing COM security levels
```

Nothing else is printed.

jviotti:
Thanks a lot @barchard. Can you try the `debug-com` branch? I added even more debugging information for that particular part of the code.

barchard:
[...]

PR #200, opened 2017-07-30
fix(windows): ignore virtual drives

jviotti:
Drivelist currently crashes if there is a virtual drive mounted. The
best solution I could find is to check the resulting error from
IOCTL_STORAGE_GET_DEVICE_NUMBER.

Change-Type: patch
Changelog-Entry: Fix "Couldn't scan the drives: An unknown error occurred" where there is a mounted virtual drive.
Fixes: https://github.com/resin-io-modules/drivelist/issues/198


src/windows/scanner.cc
```diff
@@ -186,8 +186,23 @@ ScanMountpoints(std::vector<drivelist::mountpoint_s> *const output) {
...
+    if (FAILED(result)) {
+
+      // This error, "Invalid Function", can happen when attempting
+      // to get the device number of a virtual disk.
+      // See: https://github.com/resin-io-modules/drivelist/issues/198
+      //
+      // TODO: According to the documentation, the ERROR_INVALID_FUNCTION
+      // is the constant we want here, but in practice, that constant
+      // equals "1," and not the hexadecimal number outlined in the docs.
+      // See https://msdn.microsoft.com/en-us/library/windows/desktop/dd319335(v=vs.85).aspx
```
lurch: That page specifically relates to "Portable Device" APIs, which we're not using here.
Seems like for the "standard Win32" [APIs](https://msdn.microsoft.com/en-us/library/cc231199.aspx) we're using, `ERROR_INVALID_FUNCTION` does indeed map to `0x1`, and it's just coincidence that `0x80070001` maps to `ERROR_INVALID_FUNCTION` in the "Portable Device" APIs.

I couldn't figure out what constant (if any) `0x80070001` maps to in the "standard Win32" APIs, but I think your talk of "Invalid Function" here (and in the commit message) is a bit of a red herring.
The nearest thing I could find was https://msdn.microsoft.com/en-us/library/cc704587.aspx but that lists nothing between `0x8004E031` and `0x80070005`, so perhaps `0x80070001` is simply an undocumented error code? :confused: 

All AFAICT, of course! :wink:

lurch:
...and the CI builds are all failing because of:
```
> eslint lib tests example && cpplint --recursive src
src/windows/scanner.cc:190:  Redundant blank line at the start of a code block should be deleted.  [whitespace/blank_line] [2]
src/windows/scanner.cc:195:  Missing username in TODO; it should look like "// TODO(my_username): Stuff."  [readability/todo] [2]
```

src/windows/scanner.cc
jviotti: Oh, I see. Hm, I'll at least update the comment to reflect your findings.

src/windows/scanner.cc
```diff
@@ -186,8 +186,20 @@ ScanMountpoints(std::vector<drivelist::mountpoint_s> *const output) {
     ULONG number;
     drivelist::Debug("Getting device number");
     result = drivelist::volume::GetDeviceNumber(volume, &number);
-    if (FAILED(result))
+    if (FAILED(result)) {
+
+      // This error, "Invalid Function", can happen when attempting
+      // to get the device number of a virtual disk.
```
lurch: IMHO you should get rid of the "Invalid Function" part and just change it to `This error can happen when attempting to get the device number of a virtual disk.` with a similar change also made to the commit message.

jviotti: Fixed!

jviotti:
Fixed!

lurch:
Does this also fix `SUBST` drives? https://en.wikipedia.org/wiki/SUBST
(or were they not a problem in the first place? Afraid I'm not set up to test on Windows ATM)

jviotti:
It should work fine, because under the hood, there is still just a list of disk volumes.

jhermsmeier:
What exactly qualifies as a "virtual drive"? Mounted VHDs are still (correctly so) listed, for example.

jviotti:
@jhermsmeier Do you know what are all types of virtual drives that we should test?

lurch:
@jhermsmeier I know very little about virtual drives, but see this comment: https://github.com/resin-io-modules/drivelist/issues/198#issuecomment-318591064

jhermsmeier:
Turns out, those aren't virtual drives in that sense, but user-mode file systems – like https://www.eldos.com/cbfs/ and similar. As those do not emulate block devices, they don't have a device number, afaik – hence the error.

lurch:
@jhermsmeier Ahhhh, so does http://www.fs-driver.org/ fall into that same category?

jhermsmeier:
@lurch nope, that's a kernel-mode fs driver (an actual driver), which will be handled by Windows, and has nothing to do with the underlying device.

User-mode file systems (i.e. FUSE, Dokan, Eldos CBFS, etc.) - on the contrary - work by having one common kernel-mode driver that acts as a server to user-space programs that implement the client API to that.

PR #202, opened 2017-07-31
fix(windows): don't attempt to initialize COM security levels twice

jviotti:
We use global functions to initialize COM. If the client executes
multiple scan operations at the same time on the same process, then COM
will be initialized twice, eventually resulting in an error.

The error is `RPC_E_TOO_LATE`, which according to the Windows
documentation, means that COM was already initialized. This commit
workarounds the initialization error by catching this HRESULT and
assuming everything is fine, but the correct solution is to get rid of
COM altogether.

Change-Type: patch
Changelog-Entry: Fix "Couldn't initialize the scanner: An unknown error occurred" error.
Fixes: https://github.com/resin-io-modules/drivelist/issues/198


jviotti:
@hedss Can we make VB rebase before creating the bump commit?

PR #203, opened 2017-07-31
refactor(windows): don't use COM/WMI to retrieve a drive media type

jviotti:
Turns out we already have this information in the `DISK_GEOMETRY_EX`
struct we were getting from `GetSize()`. In order to not duplicate work,
this commit refactors `GetSize()` into `GetInformation()`, which gives
back the size and now the media type removable status.

Change-Type: patch


src/windows/disk.cc
```diff
@@ -17,10 +17,16 @@
 #include "src/windows/disk.h"
 #include "src/log.h"
 
-HRESULT drivelist::disk::GetSize(std::string disk, LONGLONG *out) {
+HANDLE drivelist::disk::OpenHandle(const std::string &disk) {
   drivelist::Debug("Opening handle on " + disk);
-  HANDLE handle = CreateFile(disk.c_str(), 0, FILE_SHARE_READ, NULL,
-                             OPEN_EXISTING, 0, NULL);
+  return CreateFile(disk.c_str(), 0, FILE_SHARE_READ, NULL,
```
lurch: Does this really need a separate function? ;-)

PR #204, opened 2017-08-04
fix(windows): ensure we get real HRESULTs from failures

jviotti:
In various cases, we fallback to returning `E_HANDLE` or `E_FAIL` when
we could get the precise error code from `GetLastError()`.

This ensures we have enough information to fix bugs later on.

Change-Type: patch
Changelog-Entry: Get more detailed error codes from the Windows implementation.


jviotti:
Ping @hedss . Do you know what's going on there? The error seems to happen after the commit was merged.

PR #206, opened 2017-08-11
fix(windows): correctly handle BitLocker locked volumes

jviotti:
Locked volumes throw a FVE_E_LOCKED_VOLUME error on
GetVolumeInformation(). If this is the case, we should also set
`writable` to FALSE, since we're not able to make any changes to locked
drives (e.g. running diskpart clean on it).

See: https://github.com/resin-io/etcher/issues/1687
Change-Type: patch
Changelog-Entry: Correctly handle BitLocker locked Windows volumes.


src/windows/scanner.cc
```diff
@@ -176,11 +176,21 @@ ScanMountpoints(std::vector<drivelist::mountpoint_s> *const output) {
...
     BOOL hasFilesystem;
     drivelist::Debug("Checking if volume has filesystem");
     result = drivelist::volume::HasFileSystem(volume, &hasFilesystem);
-    if (FAILED(result))
-      return InterpretHRESULT(result);
+    if (FAILED(result)) {
+      if (result == FVE_E_LOCKED_VOLUME) {
+        drivelist::Debug("Device is locked using BitLocker");
+        hasFilesystem = FALSE;
+        // We set this drive as non-protected since if the
```
lurch: non-protected -> non-writable?

lurch:
Took me a while to figure out that FVE must mean Full Volume Encryption ;-)
And took even longer to actually find https://msdn.microsoft.com/en-us/library/windows/desktop/dd542648(v=vs.85).aspx

src/windows/scanner.cc
jviotti: Fixed!

jviotti:
Thanks for the link. I added it as a comment!

PR #209, opened 2017-08-22
fix(linux): use $XDG_RUNTIME_DIR on GNU/Linux

jviotti:
Some systems mount `/tmp` with `noexec`, causing `spawn EACCES` errors.

See: https://github.com/resin-io/etcher/issues/1699
See: https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html
Change-Type: patch
Changelog-Entry: Store GNU/Linux temporary scripts in `$XDG_RUNTIME_DIR`.


lib/execute.js
```diff
@@ -25,6 +25,13 @@ const os = require('os');
 const crypto = require('crypto');
 
 /**
+ * @summary A temporary directory to store scripts
+ * @type {String}
```
jhermsmeier: Would it be worth adding a note here on why XDG_RUNTIME_DIR is necessary?

jviotti: Done!

lurch:
Aaarrrgghhhh, I can't believe I didn't spot this before...! I must have _really_ taken my eye off the ball :-/

On Ubuntu Linux (at least in version 14.04) `$XDG_RUNTIME_DIR` defaults to something like `/run/user/1000` (where `1000` is the uid of the current user). But `/run/user/` is _itself_ mounted with the `noexec` option, which is what this PR was trying to get around in the first place!
```bash
$ mount | grep /run/user
none on /run/user type tmpfs (rw,noexec,nosuid,nodev,size=104857600,mode=0755)
```

This means the script that gets created in `$XDG_RUNTIME_DIR` can't actually be executed, and you get this in Etcher's DevTools console:
```
Uncaught Error: spawn EACCES
    at exports._errnoException (util.js:1022:11)
    at ChildProcess.spawn (internal/child_process.js:313:11)
    at exports.spawn (child_process.js:399:9)
    at Object.exports.execFile (child_process.js:160:15)
    at Object.module.(anonymous function) [as execFile] (ELECTRON_ASAR.js:200:20)
    at executeScript (/home/anndy/etcher/node_modules/drivelist/lib/execute.js:101:16)
    at createTemporaryScriptFile (/home/anndy/etcher/node_modules/drivelist/lib/execute.js:167:5)
    at fs.writeFile (/home/anndy/etcher/node_modules/drivelist/lib/execute.js:81:5)
    at FSReqWrap.oncomplete (fs.js:112:15)
```

lurch:
In conclusion, I think that this change should be reverted, else we risk breaking Etcher for all Ubuntu users.

I think ultimately the only 'proper' fix for https://github.com/resin-io/etcher/issues/1699 will be to move the drivelist-logic to native code (like we've already done for Windows), as there doesn't seem to be an 'official' way of getting a temporary directory path which is guaranteed to not have `noexec` enabled.
As a short-term stop-gap solution, I wonder if we could get away with just rewriting the bash scripts as native nodejs scripts?

lurch:
...or as an even quicker solution, we could change https://github.com/resin-io-modules/drivelist/blob/master/lib/execute.js#L102 to do:
```js
childProcess.execFile(`bash ${scriptPath}`, (error, stdout, stderr) => {
```
(or whatever the correct syntax is) instead of just `execFile(scriptPath)` directly. We _might_ need to give the full path to `bash` in this case though, and we might also need different bash-paths depending if we're running on Linux or Mac? (and this change would also make the `#!` line in the script itself redundant)

probonopd:
Which systems have mounted `/tmp` as `noexec`, since when, and why?

lurch:
@probonopd There's a bit more info in the linked issue: https://github.com/balena-io/etcher/issues/1699

Beyond that, I have no idea :wink:

Issue #215, opened 2017-09-20
lodash dependency adds 5MB to drivelist


robinwassen:
Would like to start with saying thanks for the module! Really appreciated the effort by @jhermsmeier to create native bindings rather than using the script approach for Windows, works so much better.

I started to investigate ways for us to lower the application size of our Electron-app and started by listing the largest node modules that we use.

One of those is lodash that is 5MB which is a dependency of drivelist, it is listed as a production dependency and is therefore included in the final build of the application.

lodash is used in the script output parser (parse.js) to provide 10 convenience functions when parsing the output.

**The approaches that I see would solve this problem are:**
- (quickfix) Include only the built lodash file instead of as a node module
- Replace the usage of lodash in parse.js
- Replace the javascript script output parsing by calling native modules with V8 bindnings (#172)

Would the quickfix be considered as acceptable? I think that we can reduce the lodash dependency from ~5MB to ~4KB by including the .js file instead of using the idiomatic way of including it as a node module.

robinwassen:
Performed some benchmarks on the effects of using lodash via npm or vendoring.

Command to install dependencies: `npm install --production`
Branch: master

Windows, NodeJS 6.11.3, npm 5.0.3:
- lodash via npm, total dependencies size: 4.5MB
- lodash vendored, total dependencies size:  2.0MB

macOS, NodeJS 6.11.1, npm 3.10.10
- lodash via npm, total dependencies size: 6.4MB
- lodash via vendor, total dependencies size: 2.0MB

Total dependencies size is measured by du and adding the node_modules and vendor directory sizes together.

I am not sure why the lodash file size is greater on macOS since it seems to be the identical amount of files, guess it has something to do with the file system and many small files.

mceachen:
fwiw parse.js uses 10 different lodash functions, but each one is a one-liner.

jhermsmeier:
Closed in #235

PR #216, opened 2017-09-22
Vendoring of lodash to reduce module size

robinwassen:
This to reduce the production size of the drivelist module.

See #215


Page-:
Hey, the problem of this PR it that it will actually increase the size of the drivelist module for any case where another dependency also depends on lodash as npm will not be able to dedupe lodash and so the cost will have to be paid for every module that depends on lodash.  It would be much better to use something like https://webpack.js.org/ in the end product in order to remove unused code/files, minify code, and also consolidate shared dependencies

robinwassen:
@Page- I hear you, but does webpack actually strip unused code? Otherwise an effort is needed to use the modularized lodash functions in order for that to work, that also voids the argument of deduping since that will also cause duping.

jhermsmeier:
Webpack will not work with this, as drivelist has native bindings (also, Webpack shouldn't be used to preprocess npm packages' contents) – the best solution here would be to just install the individual lodash components (every component is also [published separately on npm](https://npms.io/search?q=lodash)), which will let npm dedupe it properly while only using the necessary parts.

robinwassen:
@jhermsmeier I will give that a shot and see if it gives a positive result.

Closing this PR and creating a new one for that.

jhermsmeier:
Sweet, thanks!

Page-:
@jhermsmeier it's perfectly possible to use webpack to minify npm packages, but you're right that you need to handle native dependencies separately, by marking them as "external" so that they are still loaded via node_modules (https://github.com/resin-io/resin-supervisor/ uses webpack for minification and has native modules)

Issue #218, opened 2017-09-23
Create prebuilt binaries


jhermsmeier:

lurch:
Connects to #217

Issue #217, opened 2017-09-22
Install fails on windows


Page-:
```
gyp ERR! stack Error: Can't find Python executable "python", you can set the PYTHON env variable.
```
Is it possible to use node-pre-gyp to install pre-built binaries where possible avoiding issues where build tools are not available

jviotti:
Python is a requirement of `node-gyp` itself (see https://github.com/nodejs/node-gyp#on-windows). You can alternatively install `npm install --global --production windows-build-tools`. Would that work for you?

Page-:
@jviotti that's why I suggested node-pre-gyp, which can avoid the need for python and other build tools by making use of prebuilt binaries when available and only falling back to node-gyp if there aren't any prebuilt binaries for the current os/arch combo

jhermsmeier:
Closing this, as this has been fixed by #219 and #223

PR #219, opened 2017-09-25
feat(package): Add prebuilds for native bindings

jhermsmeier:
This adds support for prebuilding and publishing native bindings
via [prebuild] and [prebuild-install]

[prebuild]: https://github.com/prebuild/prebuild
[prebuild-install]: https://github.com/prebuild/prebuild-install

Change-Type: minor
Connects To: #218, #217


appveyor.yml
```diff
@@ -25,3 +25,10 @@ test_script:
   - node --version
   - npm --version
   - cmd: npm test
+
+deploy:
+  on:
+    branch: master
```
jhermsmeier: This stuff ^ doesn't appear to work, and their documentation doesn't really elaborate on how it'd be possible to do this (apart from that tag & branch selectors seem to be mutually exclusive)

jviotti: I contacted Appveyor about this in particular at some point. This is what we do in Etcher (adapted to your use case):

```
deploy_script:
  - if %APPVEYOR_REPO_BRANCH%==master (npm run prebuild -- -u %GITHUB_TOKEN%)
```

jhermsmeier: Ah, I see – thanks!

lurch:
Apologies for the stupid question, but how do npm / gyp / pre-gyp / prebuild / prebuild-install / etc. all inter-relate?

jhermsmeier:
> Apologies for the stupid question, but how do npm / gyp / pre-gyp / prebuild / prebuild-install / etc. all inter-relate?

npm ships with node-gyp (which is the default build system for native bindings), node-pre-gyp is node-gyp for prebuilds (which requires plenty of config & running things to produce prebuilds for all wanted combinations), and then there's prebuild (which abstracts the config & running things with node-pre-gyp away) – and prebuild-install is the slim part that'll find & download the correct prebuild binary during package installation.

And that's also why prebuild goes into dev-dependencies, and prebuild-install into the dependencies. The prebuild builds & publishes the binaries, prebuild-install installs them.

> P.S. Does this mean that Etcher builds will be faster too, as it'll now use the pre-compiled drivelist, instead of having to compile it from source?

Yup, should be a wee bit faster – our main slowdown in the realms of native bindings is building lzma-native though, I think.

lurch:
Ahhh, seems to make a bit more sense, thanks for the explanation :-)  (I didn't realise the prebuild stuff was effectively just wrappers around the node-gyp stuff)

> our main slowdown in the realms of native bindings is building lzma-native though

Could we encourage (suggest) the author of lzma-native to also use this prebuild stuff?

lurch:
Does this relate to https://github.com/resin-io/process/issues/75 at all? Or is that issue more about publishing source, whereas this PR is more about publishing binaries?

jhermsmeier:
Oh yeah, we should add that to the publishing guidelines; to have prebuilds if there are native bindings. Good catch!

Issue #222, opened 2017-09-28
Prebuilds failing on Windows


jhermsmeier:
https://ci.appveyor.com/project/resin-io/drivelist/build/1.0.830/job/k570q7x53q2v79x7

```
Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
  code.cc
  log.cc
  drivelist.cc
  com.cc
C:\Users\appveyor\.node-gyp\iojs-1.8.0\deps\v8\include\v8.h(4028): error C2610: 'v8::WasmCompiledModule::TransferrableModule::TransferrableModule(v8::WasmCompiledModule::TransferrableModule &&)' : is not a special member function which can be defaulted (..\src\drivelist.cc) [C:\projects\drivelist\build\drivelist.vcxproj]
C:\Users\appveyor\.node-gyp\iojs-1.8.0\deps\v8\include\v8.h(4031): error C2610: 'v8::WasmCompiledModule::TransferrableModule &v8::WasmCompiledModule::TransferrableModule::operator =(v8::WasmCompiledModule::TransferrableModule &&)' : is not a special member function which can be defaulted (..\src\drivelist.cc) [C:\projects\drivelist\build\drivelist.vcxproj]
C:\Users\appveyor\.node-gyp\iojs-1.8.0\deps\v8\include\v8.h(4104): error C2610: 'v8::WasmModuleObjectBuilder::WasmModuleObjectBuilder(v8::WasmModuleObjectBuilder &&)' : is not a special member function which can be defaulted (..\src\drivelist.cc) [C:\projects\drivelist\build\drivelist.vcxproj]
C:\Users\appveyor\.node-gyp\iojs-1.8.0\deps\v8\include\v8.h(4106): error C2610: 'v8::WasmModuleObjectBuilder &v8::WasmModuleObjectBuilder::operator =(v8::WasmModuleObjectBuilder &&)' : is not a special member function which can be defaulted (..\src\drivelist.cc) [C:\projects\drivelist\build\drivelist.vcxproj]
  scanner.cc
  volume.cc
  disk.cc
  wmi.cc
  win_delay_load_hook.cc
prebuild ERR! build error 
prebuild ERR! stack Error: `msbuild` failed with exit code: 1
prebuild ERR! stack     at ChildProcess.onExit (C:\projects\drivelist\node_modules\node-gyp\lib\build.js:258:23)
prebuild ERR! stack     at emitTwo (events.js:125:13)
[...]

PR #223, opened 2017-09-28
chore(appveyor): Explicitly set base image

jhermsmeier:
As Visual Studio 2013 is lacking some features that are
finding use in newer Node / V8 versions, this explicitly
specifies to use the VS2015 base image

Change-Type: patch
Connects To: #222


lurch:
Presumably this doesn't introduce any backwards-incompatibility problems?

jhermsmeier:
No, this is purely on the C++ language feature side, not the binaries themselves or anything

jhermsmeier:
Rebased against master

PR #227, opened 2017-10-02
chore(ci): Add CI deployment scripts

jhermsmeier:
This adds deployment step scripts for the prebuilds to
avoid the CIs accidentally exposing secrets through the logs.

Change-Type: patch


lurch:
I wonder if the scripts should be named something like `prebuild_publish` or `prebuild_and_publish`, to make their use more obvious? \*shrug\*

jhermsmeier:
Hmm, good point – renamed them to `prebuild-publish.*`

PR #229, opened 2017-10-06
fix: workaround noexec by executing scripts with bash

jviotti:
The temporary or `XDG_RUNTIME_DIR` directory might be mounted with the
`noexec` option, which means we might not have execution permissions to
run the scripts we just extracted.

As an easy workaround, we can call `bash` passing it the script as an
argument rather than executing the script directly.

Change-Type: patch
Changelog-Entry: Correctly execute macOS and GNU/Linux scripts inside directories mounted as `noexec`.


lurch:
Links to https://github.com/resin-io/etcher/issues/1699

PR #235, opened 2017-11-29
refactor: Remove lodash

jhermsmeier:
This removes use of lodash throughout to reduce size.

Connects-To: https://github.com/resin-io-modules/drivelist/issues/215
Change-Type: patch


lib/parse.js
```diff
@@ -58,65 +57,61 @@ const yaml = require('js-yaml');
...
+        .filter((line) => {
+          return /^(\s\s-\s)?[a-z]+:/i.test(line);
+        })
+        .map((line) => {
+          return line
+            .replace(/\\[^.\\]/g, (match, index, string) => {
+              const escapedCharacter = match[match.length - 1];
 
-          // Remove non printable ascii characters
-          // See http://stackoverflow.com/a/24229554
```
lurch: I guess there's no reason to ditch the comment?

lurch:
Failing on Node 4 - should we 'fix' that by dropping support for NodeJS version 4, now that version 8 is LTS?

jhermsmeier:
Oh, forgot about that one there; should be fixed for Node v4 now. I think we should still support Node v4 for a good while, especially since it's really not a hassle to do so.

jhermsmeier:
Hmmmpf. Travis Mac OS builds detonated with
```
The command "if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then pip install --user cpplint; else pip install cpplint; fi" failed and exited with 127 during .
```

lurch:
https://www.traviscistatus.com/ says Travis had network connectivity issues this morning...

jhermsmeier:
Grrr `/Users/travis/.travis/job_stages: line 57: pip: command not found`

lurch:
Hmmm, looking at https://github.com/resin-io/etcher/blob/master/.travis.yml and https://github.com/resin-io/etcher/blob/master/scripts/ci/install.sh it looks like `pip` _should_ be available on OSX? :confused: 

...which is what https://docs.travis-ci.com/user/reference/osx#Python-related-tools claims too :confounded:

jhermsmeier:
Yeah, it's new, and it's on Travis' side; not much we can do about it atm

lurch:
...apart from contacting their support team? ;-)
Maybe it's related to [this change](https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce) and they haven't realised there's some things still missing?

lurch:
Looks like this also needs to be rebased to master in order to pick up the fix from #239

(I wonder how many wasted developer-hours TravisCI would have saved if they'd just done `ln -s pip2 pip` in their OSX base image? :confused: )

lib/parse.js
jhermsmeier: That was unintentional. Fixed!

lurch:
LGTM (except it needs to be rebased). IMHO @jviotti should approve this.

lurch:
Cool. Fancy publishing a new version to NPM, so that we can include it in Etcher? :)

PR #236, opened 2017-11-30
fix(package): Fix prebuild script being run on build

jhermsmeier:
Change-Type: patch


lurch:
Looks like this needs to be rebased to master in order to pick up the fix from #239

PR #243, opened 2018-01-02
feat: Rewrite native bindings to use IOCTL

jhermsmeier:
This completely refactors the native bindings on Windows
to avoid COM/WMI etc. and solely rely on DeviceIoControl.
This also introduces some new properties to the device,
and fixes a range of errors / bugs.

Further changes / fixes:

- attached virtual hard drives don't get labeled as system drives
anymore
- errors with one device don't stop enumeration of other devices (and
don't cause errors to be thrown)
- info about bus / protocol (SCSI, SATA, USB, UAS, etc.)
- isReadonly is now only true if the device itself is readonly (i.e. the
switch on the sd-card adapter is flicked), not when a fs or volume is
mounted or marked as readonly
- added SD-card detection
- added detection of virtual storage devices
- BitLocker volumes no longer cause trouble, and devices containing one
will be enumerated as any other storage device
- added block size detection (both physical & logical)

**TODO:**

- [x] Adjust Linux & Mac OS scripts to yield new properties as well (defaulting to something if n/a)
- [x] Remove printf() info (translate to debug log, ideally by passing `debug` module from JS -> C++)
- [x] Update docs (README etc)
- [x] Augment tests

<details>
<summary>Example Output</summary>

```js
[{
  enumerator: 'SCSI',
  busType: 'SATA',
  busVersion: '2.0',
  device: '\\\\?\\scsi#disk&ven_wdc&prod_wd1600bevs-07rst#4&5a60d67&0&000000#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}',
  raw: '\\\\.\\PhysicalDrive0',
  description: 'WDC WD1600BEVS-07RST0',
  error: '',
  size: 160041885696,
  blockSize: 512,
  logicalBlockSize: 512,
  mountpoints: ['D:\\'],
  isReadOnly: false,
  isSystem: true,
  isVirtual: false,
  isRemovable: false,
  isCard: false,
  isSCSI: true,
  isUSB: false,
  isUAS: false
}, {
  enumerator: 'SCSI',
  busType: 'FILEBACKEDVIRTUAL',
  busVersion: '2.0',
  device: '\\\\?\\scsi#disk&ven_msft&prod_virtual_disk#2&1f4adffe&1&000003#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}',
  raw: '\\\\.\\PhysicalDrive5',
  description: 'Microsoft Virtual Disk',
  error: '',
  size: 10485760,
  blockSize: 512,
  logicalBlockSize: 512,
  mountpoints: ['I:\\'],
  isReadOnly: false,
  isSystem: false,
  isVirtual: true,
  isRemovable: true,
  isCard: false,
  isSCSI: true,
  isUSB: false,
  isUAS: false
}, {
  enumerator: 'SD',
  busType: 'SD',
  busVersion: '2.0',
  device: '\\\\?\\sd#disk&generic&sc16g&8.0#5&c518b2e&0&c3964099&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}',
  raw: '\\\\.\\PhysicalDrive4',
  description: 'Generic SC16G SD Card',
  error: '',
  size: 15931539456,
  blockSize: 4096,
  logicalBlockSize: 512,
  mountpoints: ['G:\\'],
  isReadOnly: false,
  isSystem: false,
  isVirtual: false,
  isRemovable: true,
  isCard: true,
  isSCSI: true,
  isUSB: false,
  isUAS: false
}, {
  enumerator: 'SCSI',
  busType: 'USB',
  busVersion: '2.0',
  device: '\\\\?\\scsi#disk&ven_usb3.0&prod_#000000#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}',
  raw: '\\\\.\\PhysicalDrive2',
  description: 'USB3.0  SCSI Disk Device',
  error: '',
  size: 500107862016,
  blockSize: 4096,
  logicalBlockSize: 512,
  mountpoints: ['E:\\'],
  isReadOnly: false,
  isSystem: false,
  isVirtual: false,
  isRemovable: true,
  isCard: false,
  isSCSI: true,
  isUSB: false,
  isUAS: true
}, {
  enumerator: 'SCSI',
  busType: 'SATA',
  busVersion: '2.0',
  device: '\\\\?\\scsi#disk&ven_samsung&prod_ssd_850_evo_m.2#4&5a60d67&0&020000#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}',
  raw: '\\\\.\\PhysicalDrive1',
  description: 'Samsung SSD 850 EVO M.2 120GB',
  error: '',
  size: 120034123776,
  blockSize: 512,
  logicalBlockSize: 512,
  mountpoints: ['C:\\'],
  isReadOnly: false,
  isSystem: true,
  isVirtual: false,
  isRemovable: false,
  isCard: false,
  isSCSI: true,
  isUSB: false,
  isUAS: false
}, {
  enumerator: 'USBSTOR',
  busType: 'USB',
  busVersion: '2.0',
  device: '\\\\?\\usbstor#disk&ven_disk&prod_name&rev_ax10#0012345667888&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}',
  raw: '\\\\.\\PhysicalDrive3',
  description: 'Disk Name USB Device',
  error: '',
  size: 1000204886016,
  blockSize: 512,
  logicalBlockSize: 512,
  mountpoints: ['F:\\'],
  isReadOnly: false,
  isSystem: false,
  isVirtual: false,
  isRemovable: true,
  isCard: false,
  isSCSI: false,
  isUSB: true,
  isUAS: false
}, {
  enumerator: 'SCSI',
  busType: 'FILEBACKEDVIRTUAL',
  busVersion: '2.0',
  device: '\\\\?\\scsi#disk&ven_msft&prod_virtual_disk#2&1f4adffe&1&000004#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}',
  raw: '\\\\.\\PhysicalDrive6',
  description: 'Microsoft Virtual Disk',
  error: '',
  size: 536870912,
  blockSize: 512,
  logicalBlockSize: 512,
  mountpoints: ['B:\\'],
  isReadOnly: false,
  isSystem: false,
  isVirtual: true,
  isRemovable: true,
  isCard: false,
  isSCSI: true,
  isUSB: false,
  isUAS: false
}]
```
</details>

<br>

Change-Type: major


binding.gyp
```diff
@@ -7,21 +7,29 @@
...
-        "src/log.cc"
+        "src/drivelist.cpp",
+        "src/device-descriptor.cpp",
       ],
       "conditions": [
+        [ 'OS=="mac"', {
+          "sources": [
+            "src/darwin/list.cpp"
+          ],
+          "link_settings": {
```
jviotti: I think you can remove this whole block if its not needed.

src/darwin/list.cpp
```diff
@@ -14,11 +14,15 @@
...
-#define SRC_DRIVELIST_H_
-
 #include <nan.h>
+#include "../drivelist.hpp"
+
+namespace Drivelist {
 
-NAN_METHOD(list);
+  // TODO(jhermsmeier): Implement
+  std::vector<DeviceDescriptor> ListStorageDevices() {
```
jviotti: Looks like every os `list.cpp` is the same. Can we make them all one file?

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+0x2ACCFE60L, 0xC130, 0x11D2, { 0xB0, 0x82, 0x00, 0xA0, 0xC9, 0x1E, 0xFB, 0x8B }
+};
+
+const std::vector<std::string> USB_STORAGE_DRIVERS {
+  "USBSTOR", "UASPSTOR", "VUSBSTOR",
+  "RTUSER", "CMIUCR", "EUCR",
+  "ETRONSTOR", "ASUSSTPT"
+};
+
+const std::vector<std::string> GENERIC_STORAGE_DRIVERS {
```
jviotti: I think that various of these private functions/constants should be marked as `static`

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+
+  BOOL hasFriendlyName = SetupDiGetDeviceRegistryPropertyW(
+    hDeviceInfo, &deviceInfoData, SPDRP_FRIENDLYNAME,
+    &dataType, (PBYTE) wbuffer, sizeof(wbuffer), &size);
+
+  return hasFriendlyName ? WCharToUtf8(wbuffer) : std::string("");
+}
+
+bool IsSCSIDevice(std::string enumeratorName) {
+  for (std::string driverName : GENERIC_STORAGE_DRIVERS) {
```
jviotti: Small optimization: `GENERIC_STORAGE_DRIVERS` could be a `std::set`, that has a `find()` function that runs on O(1)

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    if (enumeratorName == driverName) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+bool IsUSBDevice(std::string enumeratorName) {
+  for (std::string driverName : USB_STORAGE_DRIVERS) {
```
jviotti: Same here. I think we should use `std::set`

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+
+  BOOL hasHardwareId = SetupDiGetDeviceRegistryPropertyA(
+    hDeviceInfo, &deviceInfoData, SPDRP_HARDWAREID,
+    &dataType, (LPBYTE) buffer, sizeof(buffer), &size);
+
+  if (!hasHardwareId) {
+    return false;
+  }
+
+  // printf("SPDRP_HARDWAREID: %s\n", buffer);
```
jviotti: Should this be deleted?

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+}
+
+bool IsVirtualHardDrive(HDEVINFO hDeviceInfo, SP_DEVINFO_DATA deviceInfoData) {
+  DWORD size;
+  DWORD dataType;
+  char buffer[MAX_PATH];
+
+  ZeroMemory(&buffer, sizeof(buffer));
+
+  BOOL hasHardwareId = SetupDiGetDeviceRegistryPropertyA(
```
jviotti: This can be `const`

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    hDeviceInfo, &deviceInfoData, SPDRP_HARDWAREID,
+    &dataType, (LPBYTE) buffer, sizeof(buffer), &size);
```
jviotti: Given `dataType` and `size` is not used (I think this is also the case on similar functions on this file), could we pass `NULL` to `SetupDiGetDeviceRegistryPropertyA`?

lurch:
Only a minor aesthetic issue, but would it be possible to sort the returned list by the `raw:` property? i.e. in your example above return them in the order `['\\\\.\\PhysicalDrive0', '\\\\.\\PhysicalDrive1', '\\\\.\\PhysicalDrive2', '\\\\.\\PhysicalDrive3', '\\\\.\\PhysicalDrive4', '\\\\.\\PhysicalDrive5', '\\\\.\\PhysicalDrive6']`
(although I've not looked at the code yet so perhaps this wouldn't make sense)

lurch:
Hmmm, I've just noticed `description: 'USB3.0  SCSI Disk Device'` in your example above has `busType: 'USB'` and `busVersion: '2.0'` ?? (and in fact _every_ drive in your example above has `busVersion: '2.0'` - is that correct?)

lurch:
In previous versions of `drivelist`, "system drive" simply meant any drive that we detected as non-removable (see also #126 ), whereas in this new version we have separate properties for "system drive" and "removable"; so this definitely needs to be documented and handled with care when integrating this into Etcher.

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    if (!GetDeviceBlockSize(hPhysical, device)) {
+      errorCode = GetLastError();
+      printf("[INFO] Couldn't get block size: Error %u\n", errorCode);
+    }
+
+    BOOL isWritable = DeviceIoControl(
+      hPhysical, IOCTL_DISK_IS_WRITABLE, NULL, 0,
+      NULL, 0, &size, NULL);
+
+    device->isReadOnly = !isWritable;
```
lurch: Just for the sake of consistency, would it also be worth abstracting this out into a `GetDeviceReadOnly(hPhysical, device)` function?

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+  }
```
lurch: Is this the end of the massive for-loop started on line 472? An inline comment might be helpful.

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    device.isSCSI = IsSCSIDevice(enumeratorName);
+    device.isUSB = IsUSBDevice(enumeratorName);
+    device.isCard = device.enumerator == "SD";
+    // device.isSystem = IsSystemDevice(hDeviceInfo, deviceInfoData);
+    device.isSystem = device.isSCSI && !device.isRemovable;
+    device.isUAS = device.isSCSI && device.isRemovable &&
+      !device.isVirtual && !device.isCard;
+
+    if (GetDetailData(&device, hDeviceInfo, deviceInfoData)) {
+      device.isCard = device.busType == "SD" || device.busType == "MMC";
```
lurch: Should this be
```
device.isCard = device.isCard || device.busType == "SD" || device.busType == "MMC"
```
in case `device.enumerator == "SD" && (device.busType != "SD" && device.busType != "MMC")` ? \*shrug\*  (I don't actually know if that's possible)

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    device.enumerator = std::string(enumeratorName);
+    device.description = GetFriendlyName(hDeviceInfo, deviceInfoData);
+    device.isRemovable = IsRemovableDevice(hDeviceInfo, deviceInfoData);
+    device.isVirtual = IsVirtualHardDrive(hDeviceInfo, deviceInfoData);
+    device.isSCSI = IsSCSIDevice(enumeratorName);
+    device.isUSB = IsUSBDevice(enumeratorName);
+    device.isCard = device.enumerator == "SD";
+    // device.isSystem = IsSystemDevice(hDeviceInfo, deviceInfoData);
+    device.isSystem = device.isSCSI && !device.isRemovable;
+    device.isUAS = device.isSCSI && device.isRemovable &&
```
lurch: Hmmm, so _any_ removable SCSI drive is always UAS? Is that a safe assumption to make?

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+
+    device.enumerator = std::string(enumeratorName);
+    device.description = GetFriendlyName(hDeviceInfo, deviceInfoData);
+    device.isRemovable = IsRemovableDevice(hDeviceInfo, deviceInfoData);
+    device.isVirtual = IsVirtualHardDrive(hDeviceInfo, deviceInfoData);
+    device.isSCSI = IsSCSIDevice(enumeratorName);
+    device.isUSB = IsUSBDevice(enumeratorName);
+    device.isCard = device.enumerator == "SD";
+    // device.isSystem = IsSystemDevice(hDeviceInfo, deviceInfoData);
+    device.isSystem = device.isSCSI && !device.isRemovable;
```
lurch: Huh? What about computers still running off IDE drives? Or NVMe SSDs?

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    device.isUAS = device.isSCSI && device.isRemovable &&
+      !device.isVirtual && !device.isCard;
+
+    if (GetDetailData(&device, hDeviceInfo, deviceInfoData)) {
+      device.isCard = device.busType == "SD" || device.busType == "MMC";
+      device.isUAS = device.enumerator == "SCSI" && device.busType == "USB";
+      device.isVirtual = device.isVirtual ||
+        device.busType == "VIRTUAL" ||
+        device.busType == "FILEBACKEDVIRTUAL";
```
lurch: I don't mean to sound pessimistic, but it feels like a lot of these `isXXXX` properties are based on guesswork and heuristics, so perhaps they're not as reliable as we'd want them to be?  (unless of course they get documented as possibly not being reliable)

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    DIGCF_PRESENT | DIGCF_DEVICEINTERFACE);
+
+  if (hDeviceInfo == INVALID_HANDLE_VALUE) {
+    printf("[ERROR] Invalid DeviceInfo handle\n");
+    return deviceList;
+  }
+
+  deviceInfoData.cbSize = sizeof(SP_DEVINFO_DATA);
+
+  for (i = 0; SetupDiEnumDeviceInfo(hDeviceInfo, i, &deviceInfoData); i++) {
```
lurch: Looks a bit weird having this function-call as the middle-parameter of the for-statement?  (but maybe that's just because it's a programming style I'm not used to)

src/windows/list.cpp
jhermsmeier: They all get enumerated under the SCSI storage driver in Windows, no matter the bus type (which is why bus type & enumerator are two different things); What this essentially checks is whether it's connected to an internal bus.

src/windows/list.cpp
jhermsmeier: How is this a heuristic? If the bus type is virtual, or the hardware ID is that of a known virtual device, it is virtual. That's no guesswork, only that the previous check works without GetDetailData(). And since GetDetailData() might error for various reasons, it's of interest to check what we can beforehand, and populate the properties, to have more info for debugging.

lurch:
In the previous version of `drivelist` there was a bunch of code for checking `HasFileSystem` - I can't remember what it was used for, but would it be a useful addition to this version of `drivelist`? (although I guess that's more of a partition-thing rather than a drive-thing)

Regarding your `TODO:` items - are they things you're still going to add in this PR? If so, could you change the PR title to include "WIP" so that it doesn't get merged prematurely?

> BitLocker volumes no longer cause trouble, and devices containing one
will be enumerated as any other storage device

Although if we _can_ detect that a drive contains a BitLocker volume, it might be nice if Etcher displayed a warning-label to help prevent users accidentally overwriting it? (disclaimer: I've never used BitLocker myself)

> Change-Type: minor

Wow, really? :wink:

jhermsmeier:
> Hmmm, I've just noticed description: 'USB3.0 SCSI Disk Device' in your example above has busType: 'USB' and busVersion: '2.0' ?? (and in fact every drive in your example above has busVersion: '2.0' - is that correct?)

It's what the Windows API returns; not sure why yet – could be that it actually ended up on a 2.0 bus, or that something's reporting something wrong

jhermsmeier:
> In the previous version of drivelist there was a bunch of code for checking HasFileSystem - I can't remember what it was used for, but would it be a useful addition to this version of drivelist?

It was used to check on whether it potentially had a drive letter associated with it; this impl can get by without that.

jhermsmeier:
> Although if we can detect that a drive contains a BitLocker volume, it might be nice if Etcher displayed a warning-label to help prevent users accidentally overwriting it? (disclaimer: I've never used BitLocker myself)

This module is concerned with enumerating storage devices, not their partition schemes, contained file systems or any of that, so I don't think we should implement it for the specific case of a device containing an encrypted volume (and then also only specifically BitLocker) – that's not in the scope here.

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    // TODO(jhermsmeier): Compare against mountpoints to actually determine this
+    return true;
+  }
+
+  CoTaskMemFree(windowsPath);
+  return false;
+}
+
+std::vector<std::string> GetAvailableVolumes() {
+  DWORD logicalDrivesMask = GetLogicalDrives();
```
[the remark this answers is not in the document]
jhermsmeier: Yes. GetLogicalDriveStrings() apparently has inconsistencies in the required buffer size depending on some factors, and we also only want the letter itself, without the `:\` etc. This also makes iterating the list easier / clearer.

src/windows/list.cpp
lurch: Hmmm, so does that mean `.isSCSI` doesn't actually mean what people might expect it to mean? :-/

jhermsmeier:
> Change-Type: minor
> 
> Wow, really? 😉

Haha, yeah – when I started this it didn't break compat yet; will update to major!

src/windows/list.cpp
lurch: I wasn't specifically referring to the `.isVirtual` method, more about the `.isXXXX` properties in general, e.g.
```
device.isUAS = device.isSCSI && device.isRemovable && !device.isVirtual && !device.isCard;
```
and
```
device.isUAS = device.enumerator == "SCSI" && device.busType == "USB";
```
and the fact that `IsSCSIDevice` and `IsUSBDevice` themselves only work by attempting to match a `enumeratorName` against a (possibly incomplete) list of hardcoded strings.
But like I said, perhaps I'm being too pessimistic ;-)

src/darwin/list.cpp
jhermsmeier: The Linux & Mac OS `list.cpp` files are functional stubs, so that there's nothing that can go boom on those platforms, while also facilitating a handy starting point for their coming implementations; so I'd keep them.

src/darwin/list.cpp
```diff
@@ -14,11 +14,15 @@
  * limitations under the License.
  */
 
-#ifndef SRC_DRIVELIST_H_
-#define SRC_DRIVELIST_H_
-
 #include <nan.h>
+#include "../drivelist.hpp"
```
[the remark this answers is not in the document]
jhermsmeier: Yes, `*.h` usually denotes a C header file, and with that becoming `.hpp`, `.cc` has been changed to `.cpp` for consistency.

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    return NULL;
+  }
+
+  return str;
+}
+
+char* GetEnumeratorName(HDEVINFO hDeviceInfo, SP_DEVINFO_DATA deviceInfoData) {
+  DWORD size;
+  DWORD dataType;
+  char buffer[MAX_PATH];
```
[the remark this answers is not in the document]
jhermsmeier: MAX_PATH is a windows.h constant that denotes the maximum size of a path represented within Windows. So, yes, MAX_PATH will be big enough. Bigger is basically not an option.

Regarding the mix of A & W methods; some places don't need WideChars, like drive letters. There's just no point, because the conversion would add unnecessary complexity, and drive letters only go so far (A-Z isn't something that'd require Unicode).

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+
+  return false;
+}
+
+bool IsRemovableDevice(HDEVINFO hDeviceInfo, SP_DEVINFO_DATA deviceInfoData) {
+  DWORD size;
+  DWORD dataType;
+  DWORD result = 0;
+
+  BOOL hasRemovalPolicy = SetupDiGetDeviceRegistryProperty(
```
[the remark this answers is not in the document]
jhermsmeier: Not necessary; the switch defaults to a `false` return value

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    diskNumber = deviceNumber.DeviceNumber;
+  } else {
+    errorCode = GetLastError();
+    printf("[INFO] STORAGE_GET_DEVICE_NUMBER: Error 0x%08lX\n", errorCode);
+  }
+
+  return diskNumber;
+}
+
+void GetMountpoints(int32_t deviceNumber,
```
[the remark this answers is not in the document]
jhermsmeier: Folder-mapped volumes aren't drives or storage devices; they're just mapped paths and since we're enumerating through storage drivers here, they don't even show up (which is what we want).

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+    hPhysical, IOCTL_DISK_GET_DRIVE_GEOMETRY_EX, NULL, 0,
+    &diskGeometry, sizeof(DISK_GEOMETRY_EX), &size, NULL);
+
+  // NOTE: Another way to get the block size would be
+  // `IOCTL_STORAGE_QUERY_PROPERTY` with `STORAGE_ACCESS_ALIGNMENT_DESCRIPTOR`,
+  // which can yield more (or possibly more accurate?) info,
+  // but might not work with external HDDs/SSDs
+  if (hasDiskGeometry) {
+    device->size = diskGeometry.DiskSize.QuadPart;
+    device->blockSize = diskGeometry.Geometry.BytesPerSector;
```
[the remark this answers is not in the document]
jhermsmeier: Yes. There's no winning or losing in it though, but order. The latter will get an accurate number on the physical block size, if available (it's not available for USB devices, for example, unless their UAS)

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+  "KernSafeVirtual_________",
+  "Msft____Virtual_Disk____",
+  "VMware__VMware_Virtual_S"
+};
+
+char* WCharToUtf8(const wchar_t* wstr) {
+  char *str = NULL;
+  size_t size = WideCharToMultiByte(CP_UTF8, 0, wstr, -1, NULL, 0, NULL, NULL);
+
+  if (size <= 1) {
```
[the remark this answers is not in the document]
jhermsmeier: Because "ToMultiByte" isn't "ToSingleByte" ;P But tbh, I just nicked half of this from msdn, and don't know if that particular multibyte string is guaranteed to have at least 2 bytes even if it only encodes 1 char.

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+#include "../drivelist.hpp"
+
+namespace Drivelist {
+
+// The <ntddstor.h>, and <usbiodef.h> headers include the following
+// device interface GUIDs we're interested in;
+// @see https://docs.microsoft.com/en-us/windows-hardware/drivers/install/install-reference
+// @see https://msdn.microsoft.com/en-us/library/windows/hardware/ff541389(v=vs.85).aspx
+// To avoid cluttering the global namespace,
+// we'll just define here what we need:
```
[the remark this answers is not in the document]
jhermsmeier: Not necessarily, but moved them into a header file

src/windows/list.cpp
```diff
@@ -0,0 +1,687 @@
...
+      errorCode = GetLastError();
+      if (errorCode == ERROR_NO_MORE_ITEMS) {
+        printf("[INFO] (%i) EnumDeviceInterfaces: No more items 0x%08lX\n",
+          index, errorCode);
+        result = index != 0;
+        break;
+      } else if (errorCode != ERROR_NO_MORE_ITEMS) {
+        device->error = "SetupDiEnumDeviceInterfaces: Error " +
+          std::to_string(errorCode);
+      } else {
```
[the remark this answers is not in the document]
jhermsmeier: Yup, fixed!

src/windows/list.cpp
jhermsmeier: Yeah, -ish. If it's enumerated under the SCSI storage, but is removable it pretty much can only be USB. I bet there are cases where this breaks down, but that's what the second iteration after GetDetailDataa() on this flag is for.

src/windows/list.cpp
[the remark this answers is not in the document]
lurch: Well you're using UTF8 encoding, and that only uses a single byte to encode characters below 0x7F (i.e. 7-bit ASCII) :smiley:

src/windows/list.cpp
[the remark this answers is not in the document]
lurch: > MAX_PATH is a windows.h constant that denotes the maximum size of a path represented within Windows. So, yes, MAX_PATH will be big enough. Bigger is basically not an option.

That'd obviously be true if we're dealing with filesystem paths, but we're not dealing with filesystem paths here. However I agree that it seems very unlikely that any of the strings we're retrieving here will ever even _approach_ 260 characters, so I won't argue the point.

src/windows/list.cpp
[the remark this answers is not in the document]
lurch: I guess the only possibly-minor-inconsistency here is that if (a partition on) drive1 is mounted to a drive-letter, and a partition on drive2 is mounted to just a folder-path, then the `mountpoints` property would show drive1 as having a mountpoint, but drive2 as having no mountpoints. (so a user might assume that drive2 isn't "in use", when it actually is)

Although this isn't something I've experimented with much myself, so perhaps I'm misunderstanding some of this. I did a quick google and found e.g. https://technet.microsoft.com/en-us/library/cc753321(v=ws.11).aspx which says "assign a mount-point folder path (rather than a drive letter) to the drive".

lurch:
> > Although if we can detect that a drive contains a BitLocker volume, it might be nice if Etcher displayed a warning-label to help prevent users accidentally overwriting it? (disclaimer: I've never used BitLocker myself)

> This module is concerned with enumerating storage devices, not their partition schemes, contained file systems or any of that, so I don't think we should implement it for the specific case of a device containing an encrypted volume (and then also only specifically BitLocker) – that's not in the scope here.

But just to play devil's advocate for a moment :smiling_imp:  it _is_ concerned with trying to protect users from accidentally flashing drives that they didn't intend to - after all, isn't that what the `system` and `mountpoints` properties are for? But perhaps you're right and any BitLocker-detection should be done outside of `drivelist` and in some other module. (the only reason I even _mention_ BitLocker is because so many of our previous GH issues on Etcher mentioned it, so it's clearly used by a lot of people)

src/windows/list.cpp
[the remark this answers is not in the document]
jhermsmeier: Yeah, but it might also always have a BOM (byte-order-mark) as the first byte, but I still don't really know.

src/windows/list.cpp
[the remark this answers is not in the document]
jhermsmeier: I think you're confusing that with the subst-mapped drives many users had a problem with –
 which goes the other direction, mapping a folder -> drive letter, rather than drive -> NTFS subfolder. The latter case doesn't really matter, as the storage device will still be correctly enumerated.

src/windows/list.cpp
jhermsmeier: I don't think so – SCSI being a collection of protocols / bus types, and Windows enumerating them under the SCSI storage devices yields what you would expect it to on this platform.

jhermsmeier:
> it is concerned with trying to protect users from accidentally flashing drives that they didn't intend to - after all, isn't that what the system and mountpoints properties are for?

Nope – that's what Etcher is concerned with; this merely being the data source.

> the only reason I even mention BitLocker is because so many of our previous GH issues on Etcher mentioned it, so it's clearly used by a lot of people

They mention it, because a locked BitLocker volume was causing drivelist to blow up, because the volume itself is inaccessible in that state, and drivelist was trying to access the volume (rather than the physical device) – it was a bug, nothing more in the scope this module, in my opinion.

If we decide we need to check & handle encrypted volumes on devices, we should do that elsewhere.

src/windows/list.cpp
jhermsmeier: Done!

src/windows/list.cpp
jhermsmeier: And done!

scripts/darwin.sh
```diff
@@ -65,23 +71,28 @@ for disk in $DISKS; do
...
   then
-    echo "system: True"
+    echo "isSystem: True"
   else
-    echo "system: False"
+    echo "isSystem: False"
   fi
 
+  echo "isVirtual: null"
+  echo "isRemovable: null"
```
jviotti: Maybe we should default this to `!isSystem`?

scripts/darwin.sh
```diff
@@ -65,23 +71,28 @@ for disk in $DISKS; do
...
      echo "$mountpoints" | grep "^/$"
   then
-    echo "system: True"
+    echo "isSystem: True"
   else
-    echo "system: False"
+    echo "isSystem: False"
   fi
 
+  echo "isVirtual: null"
```
jviotti: We know for sure that the current scripts don't return any virtual drive, so I think we can safely put `false` here.

jhermsmeier: With the new APFS encrypted volumes, this actually doesn't hold up anymore (disk1 is a synthesized / virtual of disk0), so I left it "null", as we can't say for sure atm

scripts/darwin.sh
jhermsmeier: Yes, good point.

src/windows/list.cpp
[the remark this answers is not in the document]
lurch: > I think you're confusing that with the subst-mapped drives

Nope ;-)

> The latter case doesn't really matter, as the storage device will still be correctly enumerated.

...with the only minor 'wrinkle' being that (AIUI) a drive mapped to a NTFS folder won't show any entries in the `mountpoints` property, as I mentioned above. I don't disagree that it'll still be correctly enumerated.

src/windows/list.cpp
```diff
@@ -484,6 +482,7 @@ bool GetDetailData(DeviceDescriptor* device,
     }
 
     device->raw = "\\\\.\\PhysicalDrive" + std::to_string(deviceNumber);
+    device->device = device->raw;
```
lurch: If `device->device` is now the same thing as `device->raw`, do the Windows examples in the README need updating again?

tests/drivelist.spec.js
```diff
@@ -32,10 +32,27 @@ describe('Drivelist', function() {
         this.executeExtractAndRunStub = m.sinon.stub(execute, 'extractAndRun');
 
         this.executeExtractAndRunStub.withArgs(scripts.linux).yields(null, [
-          'device: "/dev/sda"',
-          'description: "My drive"',
-          'size: "15 GB"',
-          'mountpoint: "/mnt/drive"'
+          'enumerator: lsblk',
+          'busType: UNKNOWN',
```
lurch: I guess it doesn't matter, but these `lsblk` and `UNKNOWN` string values should probably be quoted? (like all the other string key-values here)

PR #249, opened 2018-01-16
fix(windows): Fix segfault on Windows 7

jhermsmeier:
This fixes a segmentation fault on Windows 7, caused by not passing `&size` to DeviceIoControl(), which resulted in access to a invalid pointer.

Closed handles are now also set to `INVALID_HANDLE_VALUE`, and some value initializations have been improved.

Further this adds compiler and linker options to aid in detecting errors like missing links to libraries and exception handling.

Change-Type: patch
Changelog-Entry: Fix crash on Windows 7


src/windows/list.cpp
```diff
@@ -226,11 +227,11 @@ int32_t GetDeviceNumber(HANDLE hDevice) {
   if (result) {
     // printf("[INFO] DeviceNumber: %i\n", deviceNumber.DeviceNumber);
     diskNumber = deviceNumber.DeviceNumber;
-  } else {
-    errorCode = GetLastError();
-    // printf("[INFO] STORAGE_GET_DEVICE_NUMBER: Error 0x%08lX\n", errorCode);
   }
 
+  // errorCode = GetLastError();
```
lurch: Even though this debugging code has been commented out, shouldn't it still be within the `else` block? (so that it's in the right place if someone chooses to uncomment it in future to help their debugging)

lurch: ...in fact, perhaps it'd be more useful to put all these `printf` calls inside some kind of `#ifdef DEBUG` section, rather than commenting them all out entirely? \*shrug\*

lurch:
> caused by not passing &size to DeviceIoControl(), which resulted in access to a invalid pointer.

Aha, https://msdn.microsoft.com/en-us/library/windows/desktop/aa363216(v=vs.85).aspx actually says `If lpOverlapped is NULL, lpBytesReturned cannot be NULL.` so I guess it's kinda surprising that this only failed on Win7 and didn't also fail on Win10.

src/windows/list.cpp
```diff
@@ -486,55 +512,76 @@ bool GetDetailData(DeviceDescriptor* device,
...
+  }  // end for (index = 0; ; index++)
 
   if (hDevice != INVALID_HANDLE_VALUE) {
     CloseHandle(hDevice);
+    hDevice = INVALID_HANDLE_VALUE;
   }
 
   if (hPhysical != INVALID_HANDLE_VALUE) {
     CloseHandle(hPhysical);
+    hPhysical = INVALID_HANDLE_VALUE;
```
lurch: Given the amount of places you do this, would it be worth creating an inline function wrapper (or macro) for this?

PR #250, opened 2018-01-16
fix(windows): Impl IsSystemDevice()

jhermsmeier:
This implements the previously unimplemented method `IsSystemDevice()`
which matches the mountpoint against the systems' root folder.

Change-Type: patch
Connects To: https://github.com/resin-io-modules/drivelist/issues/126


src/windows/list.cpp
```diff
@@ -148,21 +148,26 @@ bool IsVirtualHardDrive(HDEVINFO hDeviceInfo, SP_DEVINFO_DATA deviceInfoData) {
   return false;
 }
 
-bool IsSystemDevice(HDEVINFO hDeviceInfo, SP_DEVINFO_DATA deviceInfoData) {
+bool IsSystemDevice(HDEVINFO hDeviceInfo, DeviceDescriptor *device) {
   PWSTR windowsPath = NULL;
+  BOOL isSystem = false;
   HRESULT result = SHGetKnownFolderPath(
     FOLDERID_Windows, 0, NULL, &windowsPath);
```
lurch: As I mentioned at https://github.com/resin-io-modules/drivelist/pull/243/files#r159242773 should we also check against other common "known folder paths" ? Since some people might have their ProgramFiles or UserProfile on a disk that isn't the same as the Windows disk.

src/windows/list.cpp
```diff
@@ -148,21 +148,26 @@ bool IsVirtualHardDrive(HDEVINFO hDeviceInfo, SP_DEVINFO_DATA deviceInfoData) {
...
 
   if (result == S_OK) {
     std::string systemPath = WCharToUtf8(windowsPath);
-    CoTaskMemFree(windowsPath);
-    // printf("systemPath %s\n", systemPath.c_str());
-    // TODO(jhermsmeier): Compare against mountpoints to actually determine this
-    return true;
+    printf("systemPath %s\n", systemPath.c_str());
```
lurch: Debugging prints, which ought to be commented out? ;)

src/windows/list.cpp
```diff
@@ -579,12 +584,13 @@ std::vector<DeviceDescriptor> ListStorageDevices() {
...
     device.isCard = device.enumerator == "SD";
-    // device.isSystem = IsSystemDevice(hDeviceInfo, deviceInfoData);
-    device.isSystem = device.isSCSI && !device.isRemovable;
+    device.isSystem = !device.isRemovable &&
+      (device.enumerator == "SCSI" || device.enumerator == "IDE");
     device.isUAS = device.isSCSI && device.isRemovable &&
       !device.isVirtual && !device.isCard;
 
     if (GetDetailData(&device, hDeviceInfo, deviceInfoData)) {
+      device.isSystem = device.isSystem || IsSystemDevice(hDeviceInfo, &device);
```
lurch: Hmmm, so (to paraphrase):
```
device.isSystem = (!device.isRemovable && (device.enumerator == "SCSI" || device.enumerator == "IDE")) || IsSystemDevice(hDeviceInfo, &device);
```
Are there any situations in which the Windows drive _won't_ be a non-removable SCSI or IDE drive? :wink: 

As (I think) I mentioned in #243 we need to remember that for Linux and Mac, we effectively still have:
```
device.isSystem = !device.isRemovable;
```

jhermsmeier: > Are there any situations in which the Windows drive won't be a non-removable SCSI or IDE drive?

Yes – you can boot Windows off of pretty much anything, including those certified portable USB thumb drives.

src/windows/list.cpp
```diff
@@ -579,12 +584,13 @@ std::vector<DeviceDescriptor> ListStorageDevices() {
     device.isSCSI = IsSCSIDevice(enumeratorName);
     device.isUSB = IsUSBDevice(enumeratorName);
     device.isCard = device.enumerator == "SD";
-    // device.isSystem = IsSystemDevice(hDeviceInfo, deviceInfoData);
-    device.isSystem = device.isSCSI && !device.isRemovable;
+    device.isSystem = !device.isRemovable &&
+      (device.enumerator == "SCSI" || device.enumerator == "IDE");
```
lurch: Are the strings here the same as those listed at https://github.com/resin-io-modules/drivelist/blob/master/src/windows/list.cpp#L289 ? As that doesn't mention "IDE". Or are these something else?
Are there any enumerators other than "SCSI" and "IDE" that we should always consider as being a "system drive"?

As an aside, should we have some kind of "definition" of what exactly a "system drive" is (up until now, it has simply indicated "non removable"), with our code then making best-effort attempts to match up to that definition? IYSWIM.

src/windows/list.cpp
jhermsmeier: Yup, added!

src/windows/list.cpp
jhermsmeier: Silenced!

src/windows/list.cpp
```diff
@@ -148,21 +148,30 @@ bool IsVirtualHardDrive(HDEVINFO hDeviceInfo, SP_DEVINFO_DATA deviceInfoData) {
   return false;
 }
 
-bool IsSystemDevice(HDEVINFO hDeviceInfo, SP_DEVINFO_DATA deviceInfoData) {
+bool IsSystemDevice(HDEVINFO hDeviceInfo, DeviceDescriptor *device) {
   PWSTR windowsPath = NULL;
```
lurch: Minor nit, but perhaps this variable should be renamed to `folderPath`? (and should its declaration be moved to inside the `for` loop?)

jhermsmeier: Done

Issue #231, opened 2017-10-15
`scripts` directory should be renamed


lurch:
According to a comment on https://github.com/resin-io-modules/mountutils/pull/54 the convention is that the `scripts` directory should be used for scripts related to building / publishing the package. Therefore the scripts that we've _currently_ got in there should be moved somewhere else, and then we can move the publishing scripts out of `ci` and into `scripts`.

lurch:
Closing as invalid - there's no longer a `scripts` directory in this repo.

Issue #165, opened 2017-05-16
Only create `lib/scripts.json` at install time


lurch:
...to prevent it ever getting out of date with the platform-specific scripts.
See #164 

(of course it's possible that drivelist might get rewritten in C++ before this issue gets implemented!)

lurch:
Closing as invalid - drivelist no longer uses separate scripts.

PR #254, opened 2018-03-08
feat(lib): Use lsblk directly, parse json output

jhermsmeier:
This switches from using a bash script on linux to shelling out
to `lsblk` directly, using its `--json` option to obtain & parse
machine-readable output – removing the need for external script
files entirely.

**NOTE:** This branch is based on `diskutil-plist`, which has hence has been selected as merge-base. This will automatically resolve to `master` once #253 is merged.

Change-Type: minor
Depends on: #253 
Connects To: #231, https://github.com/resin-io-modules/drivelist/issues/165


lurch:
> removing the need for external script files entirely.

I guess the stuff in the README about external scripts & yaml needs to be updated then? :wink:

jhermsmeier:
Huh? This was unexpected...

Shou:
Yeah, that is weird, I was in the middle of reviewing too...

jhermsmeier:
I'll rebase and open a new PR; needs some more work from what it looks like anyways, as it blew on the Linux boxes on Travis – maybe `--json` isn't supported on older versions of `lsblk`? Didn't find much info as to feature history, but was apparently added in 2015.

Shou:
Here are my comments anyway :)

lurch:
> but was apparently added in 2015.

...which means it probably won't work on the Ubuntu 12.04 that I think Etcher still supports? :-/

jhermsmeier:
@lurch yeah, also didn't work on Travis at all – see #255, which also adds parsing the "old" `--pairs` output.

PR #267, opened 2018-04-09
Added a fix for diskutil returning null data after a system sleep / stand-by

grumpyoldman-io:
I added this module to my electron app and noticed a lot of errors when returning my computer from a sleep state. After some digging I found out that `globalList` in [diskutil.js (line 159)](https://github.com/resin-io-modules/drivelist/blob/c14e8850b4a28c407ddcbfeec7a8e19ec942d578/lib/diskutil.js#L153) can be `null` while no errors are given.

- [x] Tests added
- [x] Linter passes


grumpyoldman-io:
sorry for the mess, I'll reopen a proper one.

PR #270, opened 2018-04-13
fix(diskutil): Improve error handling

jhermsmeier:


lib/diskutil.js
```diff
@@ -71,12 +71,16 @@ const run = (cmd, argv, callback) => {
...
   const results = [];
+  const errors = [];
   const next = (error, result) => {
     if (result !== undefined) {
       results.push(result);
     }
-    if (error || !items.length) {
-      callback(error, results);
+    if (error !== undefined) {
+      errors.push(error || null);
```
Shou: Do we need `|| null` here given the if-check?

jhermsmeier: Oh no, of course not; updated the logic to be simpler, and more robust now

PR #294, opened 2018-08-01
Fixed linting warnings

robinwassen:
This PR fixes the eqeqeq linting warnings, also added an spec with assertions of the return value for the `require('fast-plist').parse` function since return values on zero length strings is undocumented.

`childProcess.spawn(...).on('exit')` is [documented](https://nodejs.org/api/child_process.html#child_process_event_exit) to return null or number, so the eqeqeq change is safe.


thundron:
@robinwassen I included your commit in https://github.com/balena-io-modules/drivelist/pull/312 
Closing this one and thanks for the (big) patience!

PR #295, opened 2018-08-01
ci: Add .resinci.yml

brownjohnf:
Signed-off-by: Jack Brown <jack@resin.io>
Change-Type: patch


brownjohnf:
@jhermsmeier My understanding is that we didn't need alpine support on drivelist. This .resinci.yml file will restrict what we run on.

jhermsmeier:
@brownjohnf the issue was that `pip` was missing from the alpine builds, which makes them fail on drivelist, as it uses the cpplint python package to lint the native binding code. While testing this on Alpine might not be necessary, I'd think that might impact other repos as well

.resinci.yml
```diff
@@ -0,0 +1,41 @@
...
+      node_versions:
+        - "10"
+        - "8"
+        - "6"
+    - name: linux
+      os: ubuntu
+      architecture: x86
+      node_versions:
+        - "8"
+        - "6"
```
jhermsmeier: Why no Node v10 on x86? Just to reduce build count, or is there another reason?

.resinci.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: darwin
+      os: macos
+      architecture: x86_64
+      node_versions:
+        - "6"
```
jhermsmeier: Should we test Node v8 and v10 on Mac OS & Windows as well?

.resinci.yml
brownjohnf: Node v10 doesn't support x86. https://github.com/nodejs/node/issues/20751

.resinci.yml
brownjohnf: Eventually, yes, but I think we'll need to get things containerized first. Windows in particular will be a nightmare if we don't.

jhermsmeier:
@resin-ci retest

PR #296, opened 2018-08-02
Mac native bindings

robinwassen:
Rewrote the macOS implementation to be based on native calls to the DiskArbitration API:s rather than having to execute diskutil in a subprocess.

This improved the drivelist.list() function performance from ~300ms to ~20ms.

I strived to keep API compatibility, but there are a few differences compared to the diskutil implementation.

**API differences:**

- `enumerator` value has changed from diskutil to NSFileManager
- null values are now the default values such as false and empty string (I assume this is similar to how it works in the Windows implementation)
- `isVirtual` implementation is not as sophisticated as before. It only checks if the disk is "synthesized" and I think it's not as good as the VirtualOrPhysical value used before, but I have not managed to dig out the exact details how that value works under the hood in diskutil.
- `label` on mount points is missing since `DeviceDescriptor` binding does currently not support `label`

**Output comparison:**

output_compare

**Left to do:**
- Add `label` to `DeviceDescriptor`
- See if the `isVirtual` check can be improved
- Do some proper testing (I have only done limited testing during development)

Any feedback is appreciated. :)


src/darwin/list.mm
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright 2017 resin.io
```
lurch: Should this be `Copyright 2018 Robin Andersson` ? :man_shrugging:

robinwassen: @lurch Had to read up about it, and since there is no CTA for this project I think the correct would be to have multiple Copyrights: https://opensource.stackexchange.com/questions/886/copyright-and-contributing-to-an-open-source-project

Not that picky about that.

src/darwin/list.mm
```diff
@@ -0,0 +1,139 @@
...
+  NSNumber *DictNum(CFDictionaryRef dict, const void *key) {
+      return (NSNumber*)CFDictionaryGetValue(dict, key);
+  }
+
+  std::vector<DeviceDescriptor> ListStorageDevices() {
+    std::vector<DeviceDescriptor> deviceList;
+
+    NSArray* volumeKeys = [NSArray arrayWithObjects:NSURLVolumeNameKey, NSURLVolumeIsRemovableKey, nil];
+    NSArray* volumePaths = [[NSFileManager defaultManager]
+                            mountedVolumeURLsIncludingResourceValuesForKeys:volumeKeys
```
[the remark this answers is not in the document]
robinwassen: @lurch I think that you might be right, I will do some testing with unmounted storage devices and get back with the results.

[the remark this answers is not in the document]
robinwassen: Do not have access to a storage device that I can wipe yet, but I experimented a bit with using DiskArbitration to list the devices from a disk perspective instead of using the volume approach.

The good thing is that it became a bit more natural since the DeviceDescriptor will be built on a device level and does not need to be merged, instead mount points will be added to the devices afterwards.

The ugly part is that it involved some black magic to force DiskArbitration to run sync, but I can abstract that away.

jhermsmeier:
First of all, this is absolutely great, thank you! We've been planning to do this, but never ended up having enough time.

> null values are now the default values such as false and empty string (I assume this is similar to how it works in the Windows implementation)

We use `null` on purpose for values that cannot be determined (and thus not relied upon), i.e. when defaulting `isUAS` to `false`, that would imply that it's definitely not USB attached SCSI, which would be incorrect, and hence those are defaulted to `null` – of course there are exceptions, such as the `description`, which we try to synthesize as best as we can, or fall back to the device path itself, as not having a description makes it difficult to display & identity in a UI

robinwassen:
@jhermsmeier Thanks! I will look how I can solve the null values, looked at the Windows implementation and now I understand why the bool null problem wasn't present there. It does not need to return any nulls for the bools.

robinwassen:
I have solved the null values on the unused fields, changed the output to be based on disks rather than volumes, added labels to the mounts.

Also tested with a SD card attached to my computer, seems to work well both when its partitioned and when it is wiped by `diskutil zeroDisk /dev/diskX`.

The outstanding points I have are:
- If the `isVirtual` is good enough
- How I am supposed to test if this implementation holds up
- Clean up of the old implementation

The current state is that the output between the old version and new version is identical on my machine (except the enumerator field for obvious reasons).

compare_output_diskarb

jhermsmeier:
@resin-io-modules-versionbot sshhhhh! 🤐

jhermsmeier:
> I have solved the null values on the unused fields, changed the listing to be based on disks rather than volumes, added labels to the mounts.

Neat!

> How I am supposed to test if this implementation holds up

Well, that's the tricky part with the native implementations; we test these manually, reproducing rigs that caused issues, and testing in VMs – so we'll try and do that with this as well (so review might take a bit of time).

> If the isVirtual is good enough

Haven't had time to read through the code enough to be able to tell yet, but will do so in the next few days!

lurch:
> @resin-io-modules-versionbot sshhhhh! :zipper_mouth_face:

Isn't that what the `procbots/versionbot/no-checks` label is for? :wink:

robinwassen:
@jhermsmeier Added some additional changes. I brought the implementation up-to-speed with diskutil on master by adding the `devicePath` and check removability by checking `isEjectable`.

Also implemented the `isCard` field since I tripped over how-to. :)

src/darwin/list.mm
```diff
@@ -0,0 +1,155 @@
...
+    CFDictionaryRef mediaIconDict = (CFDictionaryRef)CFDictionaryGetValue(diskDescription, kDADiskDescriptionMediaIconKey);
+    NSString * iconFileName = (NSString*)CFDictionaryGetValue(mediaIconDict,
+                                                              CFStringCreateWithCString(NULL, "IOBundleResourceFile", kCFStringEncodingUTF8));
+
+    return [iconFileName isEqualToString:@"SD.icns"];
+  }
+
+
+  NSNumber *DictNum(CFDictionaryRef dict, const void *key) {
+      return (NSNumber*)CFDictionaryGetValue(dict, key);
```
jhermsmeier: Like here, there are some inconsistencies in indentation – can you normalize to 2 space indentation (since that's what we're using in this repo)?

robinwassen:
@zvin Fixed the feedback that @jhermsmeier had on this in PR #317. 

Closing this PR.

Issue #274, opened 2018-04-26
No x86 prebuilds


lurch:
For reference, see https://github.com/resin-io-modules/mountutils/issues/62
(I'm now creating a separate issue on the `drivelist` repo, because it sees much more activity than the `mountutils` repo :wink:  )

https://github.com/resin-io-modules/drivelist/releases only contains x64 prebuilds, but I _think_ it ought to also contain x86 prebuilds?

PR #299, opened 2018-08-07
fix(ci): Added x86 builds to AppVeyor

robinwassen:
Should solve #274.

Seems like AppVeyor has no problems with building x86.

Worth mentioning is that prebuilding on Windows on Node 10 will work properly now since I went in and fixed: https://github.com/prebuild/prebuild/issues/221


jhermsmeier:
@resin-ci retest

nazrhom:
@resin-ci retest

Issue #303, opened 2018-08-31
SD Card is not removable UBUNTU 18.04


amdomanska:
System: Ubuntu 18.04

When trying to flash os img to my sb card with Etcher, I've got SD card inserted and my system sees that but Etcher claims that "no removable drive detected".

running example/ gives an output:

```
[ { enumerator: 'lsblk:json',
    busType: 'SATA',
    busVersion: null,
    device: '/dev/sda',
    devicePath: 'pci-0000:00:17.0-ata-2',
    raw: '/dev/sda',
    description: 'ATA SanDisk X600 2.5 (ESP, OS, UBUNTU, [SWAP])',
    error: null,
    size: 512110190592,
    blockSize: 512,
    logicalBlockSize: 512,
    mountpoints: 
     [ { path: '/boot/efi', label: 'ESP' },
       { path: '/', label: 'UBUNTU' },
       { path: '[SWAP]', label: null } ],
    isReadOnly: false,
    isSystem: true,
    isVirtual: false,
    isRemovable: false,
    isCard: null,
    isSCSI: true,
    isUSB: false,
    isUAS: null },
  { enumerator: 'lsblk:json',
    busType: 'UNKNOWN',
    busVersion: null,
    device: '/dev/mmcblk0',
    devicePath: 'pci-0000:03:00.0-platform-rtsx_pci_sdmmc.0',
    raw: '/dev/mmcblk0',
    description: '(resin-boot, resin-rootA, resin-rootB, resin-state, resin-data)',
    error: null,
    size: 31914983424,
    blockSize: 512,
    logicalBlockSize: 512,
    mountpoints: [],
    isReadOnly: false,
    isSystem: true,
    isVirtual: false,
    isRemovable: false,
    isCard: null,
    isSCSI: null,
    isUSB: null,
    isUAS: null } ]
```
lsblk output:

```
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
loop0         7:0    0   2.3M  1 loop /snap/gnome-calculator/199
loop1         7:1    0 140.9M  1 loop /snap/gnome-3-26-1604/70
loop2         7:2    0  14.5M  1 loop /snap/gnome-logs/40
loop3         7:3    0  34.7M  1 loop /snap/gtk-common-themes/319
loop4         7:4    0  87.9M  1 loop /snap/core/5328
loop5         7:5    0   3.7M  1 loop /snap/gnome-system-monitor/54
[...]

PR #304, opened 2018-08-31
lsblk: Check `hotplug` to set `isRemovable`

jhermsmeier:
This adds a check for the `hotplug` property in `lsblk`'s output, which
fixes SD-cards not being detected as removable on some Linux distros

Change-type: patch
Connects to: #303
Signed-off-by: Jonas Hermsmeier <jhermsmeier@gmail.com>


amdomanska:
LGTM

PR #312, opened 2019-01-10
Fix devices mountpoints for lsblk in pairs mode

thundron:
Switched from checking the Major ID (which can be the same across different devices) to checking the parent and child names

Change-type: patch
Signed-off-by: Lorenzo Alberto Maria Ambrosi <lorenzoa@balena.io>


tests/drivelist.spec.js
```diff
@@ -58,7 +58,7 @@ describe('Drivelist', function() {
             `Invalid error: ${device.error}`
           );
           assert.ok(
-            Number.isFinite(device.size),
+            Number.isFinite(device.size) || device.size === null,
```
lurch: Would it make sense to swap it round to `device.size === null || Number.isFinite(device.size)` ?

thundron: @lurch I mean, it would still behave the same and they're just tests, but it looks cleaner so I'ma change it in a sec (and a separate PR for tests only)

PR #321, opened 2019-01-31
Fix lsblk JSON output utf-8 encoded characters

thundron:
Replaces https://github.com/balena-io-modules/drivelist/pull/320

Change-type: patch
Signed-off-by: Lorenzo Alberto Maria Ambrosi <lorenzothunder.ambrosi@gmail.com>

### DIFF
--- a/lib/lsblk/index.js
+++ b/lib/lsblk/index.js
@@ -80,11 +80,16 @@ const lsblk = (callback) => {
 
     let devices = null;
 
-    try {
-      devices = SUPPORTS_JSON ? parseJSON(stdout) : parsePairs(stdout);
-    } catch (e) {
-      callback(e);
-      return;
+    if (SUPPORTS_JSON) {
+      try {
+        devices = parseJSON(stdout);
+      } catch (e) {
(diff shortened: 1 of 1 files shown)


lurch:
Do you need to catch any errors from parsePairs ?

thundron:
@lurch It seems like we don't, the parsePairs function is just filtering and mapping so it won't break

lib/lsblk/index.js
```diff
@@ -80,11 +80,15 @@ const lsblk = (callback) => {
...
-      devices = SUPPORTS_JSON ? parseJSON(stdout) : parsePairs(stdout);
-    } catch (e) {
-      callback(e);
-      return;
+    if (SUPPORTS_JSON) {
+      try {
+        devices = parseJSON(stdout);
+      } catch (e) {
+        SUPPORTS_JSON = false;
+        devices = parsePairs(stdout);
```
zvin: This won't work, stdout will be some malformed json here, you can't parse it with `parsePairs`.

PR #325, opened 2019-02-20
Fix parsing lsblk --paris output for disks with no partition table

zvin:
Change-type: patch


lurch:
Hopefully this'll make it into https://github.com/balena-io/etcher/pull/2635 ? :wink:

zvin:
@lurch it will :)

PR #328, opened 2019-02-25
Update example in readme

zvin:
Change-type: patch

### DIFF
--- a/doc/README.hbs
+++ b/doc/README.hbs
@@ -29,12 +29,8 @@ Examples (the output will vary depending on your machine):
 ```js
 const drivelist = require('drivelist');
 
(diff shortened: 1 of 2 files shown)


doc/README.hbs
```diff
@@ -29,12 +29,8 @@ Examples (the output will vary depending on your machine):
...
 
-drivelist.list((error, drives) => {
-  if (error) {
-    throw error;
-  }
-
-  console.log(drives);
+const drives = await drivelist.list();
+console.log(drives);
 });
```
lurch: @zvin Should this line have been deleted too? (given that a callback function is no longer being created)

PR #330, opened 2019-02-27
Fix macos crash

### LINKED ISSUE #329: Segfault when unplugging USB stick on mac os
If I unplug while it is running, it crashes the whole Node.js process. To reproduce, run this loop:
```
(async () => {
  while (true) {
    console.log((await require('drivelist').list()).length);
  }
})();
```

Then unplug a connected USB stick

zrzka:
It fixes #329. I modified the test file a bit (attached at the end of this message). This PR:

* Fixes macOS crash when SD card is ejected
* Moves `DeviceDescription` creation to a separate function
* Adds couple of missing `CFRelease` calls (memory leaks)
* Makes `Drivelist::IsCard` more robust

`index.ts` I was using for tests:

```typescript
import { inspect } from 'util';
import { list } from '..';

async function main() {
  var last_count = -1;
  while (true) {
    let drives;
    try {
      drives = await list();
    } catch (error) {
      console.error(error);
      process.exitCode = 1;
      return;
    }

    if (last_count != drives.length) {
      last_count = drives.length;
      console.log(last_count);
    }
  }
}

main();
```

### DIFF
--- a/src/darwin/list.mm
+++ b/src/darwin/list.mm
@@ -130,6 +130,7 @@ namespace Drivelist {
       deviceList.push_back(device);
 
       CFRelease(diskDescription);
+      CFRelease(disk);
     }
 
     // Add mount points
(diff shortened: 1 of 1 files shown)


zrzka:
Yep, it just contains Change-type for the changelog.

zrzka:
@resin-ci retest

PR #333, opened 2019-03-04
Fix isVirtual value for DMG & TimeMachine on macOS

zrzka:
`drivelist` on macOS sets `isVirtual` flag to `true` if:

* `kDADiskDescriptionDeviceProtocolKey` equals to `Virtual Interface` ~~**OR**~~
* ~~`kDADiskDescriptionDeviceModelKey` equals to `Disk Image`.~~

~~It's probably enough to use just one of these, but I've decided to use both of them. It seems that this is constantly changing between macOS releases. It should still work if Apple decides to change value of one of those keys. If you'd like to use just one, let me know and I'll remove the other.~~

Change-type: patch
Signed-off-by: Robert Vojta <robert@balena.io>

### DIFF
--- a/src/darwin/list.mm
+++ b/src/darwin/list.mm
@@ -56,7 +56,6 @@ namespace Drivelist {
 
   DeviceDescriptor CreateDeviceDescriptorFromDiskDescription(std::string diskBsdName, CFDictionaryRef diskDescription) {
     NSString *deviceProtocol = (NSString*)CFDictionaryGetValue(diskDescription, kDADiskDescriptionDeviceProtocolKey);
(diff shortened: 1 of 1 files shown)


zvin:
`kDADiskDescriptionDeviceProtocolKey == "Virtual Interface"` hsould be enough I think. Was it different in older macos releases?

zrzka:
@zvin okay, I removed `kDADiskDescriptionDeviceModelKey` check, just `kDADiskDescriptionDeviceProtocolKey` remains.

I don't have older macOS release around to check. But something in my head says that there was some change, but I can be wrong. Don't have any facts / checks around now. I stopped developing macOS apps like 2-3 years ago and thus I stopped following all these changes as well.

I think it's ok as it is now and if there will be any change, it can be quickly fixed.

PR #335, opened 2019-03-08
Expose types when this module is imported

telphan:
Change-type: patch
Signed-off-by: Theodor Gherzan <theodor@resin.io>


telphan:
Not needed. Closing.

PR #336, opened 2019-03-18
Fix DASession and NSMutableArray leaks on macOS

zvin:
Change-type: patch

### DIFF
--- a/src/darwin/list.mm
+++ b/src/darwin/list.mm
@@ -101,13 +101,13 @@ namespace Drivelist {
 
   std::vector<DeviceDescriptor> ListStorageDevices() {
     std::vector<DeviceDescriptor> deviceList;
-    REDiskList *dl = [[REDiskList alloc] init];
 
     DASessionRef session = DASessionCreate(kCFAllocatorDefault);
     if (session == nil) {
       return deviceList;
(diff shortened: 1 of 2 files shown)


src/darwin/list.mm
```diff
@@ -131,6 +131,7 @@ DeviceDescriptor CreateDeviceDescriptorFromDiskDescription(std::string diskBsdNa
       CFRelease(diskDescription);
       CFRelease(disk);
     }
+    [dl dealloc];
```
zrzka: You're not supposed to call `dealloc` - never. The only place where you can do this is inside `dealloc` method where you must call `[super dealloc]` as the last thing. Should be `release`.

zrzka:
Missing `[dl release]` [here](https://github.com/balena-io-modules/drivelist/blob/929fd6110d1f3eb9bf7bdcc3d27b6fd02f5c6a93/src/darwin/list.mm#L108).

src/darwin/list.mm
zvin: amended

zvin:
>Missing [dl release] here.
@zrzka moved the init after `DASessionCreate` instead.

PR #339, opened 2019-04-09
Fix Example in Readme.md

wzr1337:
remove trailing });


thundron:
@wzr1337 You need to rewrite commit body, something like\
```
Fix Example in Readme.md

Change-type: patch
```
Other than that (required for CI checks), lgtm

lurch:
Fixed by #350

PR #351, opened 2019-10-22
Fix Example in Readme.md

armatures:
Looks like I'm the third person here to try to run the example in the README and run into this problem. I saw the commit messages need to be formatted in a specific way. Maybe this one is?

I also wrapped the example in an `async` function so you can run it as written.


lurch:
Fixed by #350

PR #353, opened 2019-11-06
Avoid segfaulting on mac when some disk description fields are null

zvin:
Change-type: patch


lurch:
Relates to https://github.com/balena-io/etcher/issues/2687#issuecomment-550292865 ?

zvin:
@lurch yes

PR #362, opened 2020-02-18
Fix lsblk --pairs parsing when a value has hexadecimal escapes

zvin:
Change-type: patch


lurch:
Seems that this is related to #320 ?

In python3 :
```python3
>>> b'\xe2\x88\x95dev\xe2\x88\x95sdi'.decode('utf8')
'∕dev∕sdi'
```

Although note that the slashes in that output string are not "regular" slashes but some weird variant...
```python3
>>> b'\xe2\x88\x95dev\xe2\x88\x95sdi'.decode('utf8').encode('utf8')
b'\xe2\x88\x95dev\xe2\x88\x95sdi'
>>> '/dev/sdi'.encode('utf8')
b'/dev/sdi'
```

PR #365, opened 2020-03-04
Modify example code to work out of the box

camow7:
When testing the old example code I would get the following error:

const drives = await drivelist.list();
               ^^^^^
SyntaxError: await is only valid in async function

I have modified the example code so you can cut, paste and run without error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,12 @@ Examples (the output will vary depending on your machine):
 ```js
 const drivelist = require('drivelist');
 
-const drives = await drivelist.list();
-console.log(drives);
+getDrives();
+
+async function getDrives () {
+    const drives = await drivelist.list();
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -176,10 +180,14 @@ Documentation
...
-drives.forEach((drive) => {
-  console.log(drive);
-});
+getDrives();
+
+async function getDrives () {
+    const drives = await drivelist.list();
+    drives.forEach((drive) => {
+        console.log(drive);
+      });
```
zvin, line 189: Please fix the indentation.
You also need to add a footer to your commit message:
```

Change-type: patch
```

tejadong: It is not a failure. You just have to add async to the function callback ...

PR #369, opened 2020-05-29
Rename resin -> balena

zvin:
Changelog-entry: Rename resin -> balena
Change-type: patch


zvin:
@balena-ci I self-certify!

PR #370, opened 2020-05-29
Remove gitter link

zvin:
Change-type: patch


zvin:
@balena-ci I self-certify!

PR #371, opened 2020-08-06
Mark APFS media as virtual and assign its mountpoints to the source drive

zvin:
Change-type: minor


zvin:
@balena-ci retest

zvin:
@balena-ci I self-certify!

PR #377, opened 2020-11-19
Include partition table info

zvin:


zvin:
@balena-ci I self-certify!

PR #380, opened 2020-11-30
Add versionbot changelog

zvin:


zvin:
@balena-ci I self-certify!

PR #383, opened 2021-01-07
Pass strings between methods as std::string instead of char *

### LINKED ISSUE #381: "basic_string::_M_construct null not valid" exception, in cases WCharToUtf8() returns NULL
It seems you let WCharToUtf8() return NULL when the input string is zero length: https://github.com/balena-io-modules/drivelist/blob/master/src/windows/list.cpp#L43

This causes problems because the result of WCharToUtf8() is fed to a std::string constructor at several places.
Here: https://github.com/balena-io-modules/drivelist/blob/master/src/windows/list.cpp#L84
and here: https://github.com/balena-io-modules/drivelist/blob/master/src/windows/list.cpp#L160

The std::string constructor does not like being fed NULL, and throws a "basic_string::_M_construct null not valid" exception in such cases. (at least with mingw/gnu c++ library. not sure what visual c++'s library does).
If exceptions are not caught by a try/catch construct either somewhere in the application, this will effectively terminate the application.

BTW WCharToUtf8() also seems to allocate memory with calloc() that does not seem to be freed anywhere.
(Note that if you pass a char* to a std::string constructor it will make a copy of the contents. It does not take over the ownership of the original memory or anything)

### LINKED ISSUE #382: windows: GetEnumeratorName() returns pointer to stack
https://github.com/balena-io-modules/drivelist/blob/master/src/windows/list.cpp#L71

In GetEnumeratorName() you create a char array variable "buffer" on the stack.
What you put on the stack is only guaranteed to be there during the scope of the function.
At the end of the function you return a pointer to said variable.

That does not look quite right...
Suggest you let your functions return std::string instead of "char *"

maxnet:
- Fixes "basic_string::_M_construct null not valid" exception
  aborting program, because WCharToUtf8() returned NULL
  in some cases, and NULL was being fed to string constructor.
- Fixes memory leak because memory allocated with calloc()
  in WCharToUtf8() was not being freed anywhere
- Fixes undefined behavior because GetEnumeratorName() returns
  pointer to stack memory, that goes outside of scope while
  pointer still is being used.

Closes #381
Closes #382

Tested with mingw/GNU libstc++ (not with vc++)

### DIFF
--- a/src/windows/list.cpp
+++ b/src/windows/list.cpp
@@ -28,6 +28,7 @@
 #include <stdint.h>
 #include <inttypes.h>
 #include <nan.h>
+#include <wchar.h>
 #include <string>
 #include <vector>
 #include <set>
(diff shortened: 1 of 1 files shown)


zvin:
Replaced with #387 
Thanks for the PR @maxnet

PR #386, opened 2021-01-19
Support lsblk versions that do no support the pttype column

zvin:
Change-type: patch


zvin:
@balena-ci retest

zvin:
@balena-ci I self-certify!

PR #392, opened 2021-07-21
ported to Node-API

TanninOne:
This ports the library to node-api.

Only tested on windows.


zwhitchcox:
Hi, I'd like to merge your PR, but we don't directly change the version. Instead, we add

```

Change-type: major
```

to the bottom of one of your commit messages. Can you please add this?

js/index.d.ts
```diff
@@ -0,0 +1,39 @@
+export interface Mountpoint {
```
zwhitchcox, line 1: This directory should not be included.

zwhitchcox (changes requested):

zwhitchcox:
Closed in favor of #399

PR #399, opened 2022-06-15
Switching to Napi

zwhitchcox:
This change will make this package compatible with all future versions of Node due to its stable ABI.

Some packages are also upgraded in this commit.

Change-type: major


zwhitchcox:
@balena-ci I self-certify!

PR #408, opened 2022-10-25
Switch to Flowzone

ab77:
Change-type: patch


ab77:
@zwhitchcox can you please glance over this - it builds/tests on all platforms.

ab77:
@zwhitchcox we can probably delete all of these also: https://github.com/balena-io-modules/drivelist/releases

zwhitchcox:
Nice! LGTM @ab77

zwhitchcox:
Yes, probably no one is depending on those...I actually do want to try to prebuild at some point, just because it makes things a lot simpler for people, but I'm not sure how that works with flowzone

zwhitchcox (approved):

PR #411, opened 2022-11-03
support node 14-18

zwhitchcox:
Updates dependencies and supports node 14-18


zwhitchcox:
@balena-ci I self-certify!

PR #416, opened 2023-01-03
Update dependencies

ab77:


ab77:
@balena-ci I self-certify!

PR #421, opened 2023-01-25
Add attachTimestamp field

MrSmoer:
I thought it'd be useful to add field which holds the unix Timestamp the drive was attached, to easily allow users to identify the right device they just plugged in.

I am new with C++ and Typescript, so please review my changes.

I don't know how to deal with the y2038 bug in C++ for 32-bit systems, so please look over this.

This is tested on Linux, Windows and MacOS 64-Bit and I tried to run the linter, but I am not shure if cpplint worked properly and ran prettier, which corrected a few typos in some other files as well.
For the general approach it uses the modifcation Timestamp of the device files under /dev on Mac and Linux, and the "Last Arrival Date" Field on Windows. This is poorly documented, but I found this in [devpkey.h](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.16299.0/shared/devpkey.h#L174). 
I figured that this is not included in mingw's devpkey.h, because it didn't work without.

Actually I don't know what I'm doing with C++ and based my changes on suggestions from stackoverflow and chatgpt.

I would really appreciate your feedback,
Have a nice Day!


MrSmoer:
Does it has to pass the workflow fist? Do I need to bump a Version number somewhere? Do you maybe have feedback on this?

lurch:
> Do you maybe have feedback on this?

Unfortunately I think that the number of open, unanswered issues https://github.com/balena-io-modules/drivelist/issues shows that this repo isn't very high on Balena's priority list? :shrug:   (of course I'd be happy to be proved wrong!)

PR #450, opened 2024-02-09
bump node-gyp version to 10.0.1

Taujor:
For compatibility with Python v3.12 and beyond since distutils has been "removed from the standard library". [Python 3.12 Changelog](https://docs.python.org/3/whatsnew/3.12.html#summary-release-highlights)


otaviojacobi:
Hello @Taujor thank you for your contribution. Sadly CI was broken aswell, which also required a small fix in order to run the tests: bump was done in #453

